Issue #22, opened 2020-01-17
depends_on cannot mapping directly to kubernetes


hangyan:
**What is the problem you're trying to solve**
depends_on cannot mapping directly to kubernetes

**Describe the solution you'd like**
mark it as optional

**Additional context**
https://github.com/kubernetes/kompose/issues/1053
https://github.com/docker/compose/issues/6855

PR #24, opened 2020-01-17
Define loose version and validation rules

### LINKED ISSUE #13: Don't require version
Better than being version-based, Compose spec should be feature driven

As my compose file use some xyz attribute, I should not have to specify a version attribute, but just let the Compose implementation tell me this attribute is not supported.

ndeloof:

Change the versioning logic so that a Compose implementation validates a Compose file regarding the features it supports, not the version used by the Compose file.
This makes `version` useless (but valid for compatiblity)

Main idea is that user can pick-up any sample found on stackoverflow and run them with his Compose tools, as long as the latter do support the fields declared in the compose file.

Other idea is that a Compose implementation MAY not support some fields, which are irrelevant in his context (see #22). This won't prevent such a tool to be considered compliant with the specification. We just REQUIRE it warn users (at least) when such unsupported fields are in use.

Fixes #13

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -9,11 +9,21 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 ## Version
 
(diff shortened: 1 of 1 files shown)


ndeloof:
Has been replaced by https://github.com/compose-spec/compose-spec/pull/82

PR #58, opened 2020-04-10
Update "host or none" section of the documentation

### LINKED ISSUE #41: Define "host or none"
Clarify the "host or none" section. 

Literally a copy/paste of what's in the [Compose file version 3 reference](https://docs.docker.com/compose/compose-file/) for "host or none" would be a great addition/replacement.

bruno-oliveira:

This serves to update the spec section related to "host or none" to match the current documentation available in the current documentation of compose.

Fixes #41 

Signed-off-by: Bruno Oliveira <olivbruno8@gmail.com>

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1244,13 +1244,14 @@ Default and available values are platform specific. Compose specification MUST s
 
 #### host or none
 
-The syntax for using built-in networks such as `host` and `none` is different, as such networks implicitly exists outside
-the scope of the Compose implementation. To use them one MUST define an external network with the name `host` or `none` and
-an alias that the Compose implementation can use (`hostnet` or `nonet` in the following examples), then grant the service
-access to that network using its alias.
+Use the host’s networking stack, or no networking. Equivalent to `docker run --net=host` or `docker run --net=none`. Only used if you use `docker stack` commands. If you use the `docker-compose` command, use network_mode instead.
+
+If you want to use a particular network on a common build, use [network](#network_mode) as mentioned in the second yaml file example.
+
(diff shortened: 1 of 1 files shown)


bruno-oliveira:
Any updates? :)

spec.md
```diff
@@ -1244,13 +1244,14 @@ Default and available values are platform specific. Compose specification MUST s
 
 #### host or none
 
-The syntax for using built-in networks such as `host` and `none` is different, as such networks implicitly exists outside
-the scope of the Compose implementation. To use them one MUST define an external network with the name `host` or `none` and
-an alias that the Compose implementation can use (`hostnet` or `nonet` in the following examples), then grant the service
-access to that network using its alias.
+Use the host’s networking stack, or no networking. Equivalent to `docker run --net=host` or `docker run --net=none`. Only used if you use `docker stack` commands. If you use the `docker-compose` command, use network_mode instead.
```
chris-crone, line 1247: We'd like to make the spec vendor neutral and would prefer to explain things in generic terms rather than just what the current Docker behaviour is.

bruno-oliveira: I see! Any suggestions on how to approach this? So idea is to explain these modes without mentioning docker?

chris-crone: Networking is going to be a tricky topic as Docker networks do not really map to other platforms. I think this is something that we're going to have to tackle with the community in the fortnightly community meetings.

You can find details about the community meeting on our Slack channel (#compose-specification) on the CNCF Slack (https://slack.cncf.io).

Toasterson: In the case of illumos there is the presence of a shared or an exclusive ip stack. You can read the manual section here https://illumos.org/man/5/zones for inspiration. It might help a lot to read how different vendors implement this kind of feature.

PR #59, opened 2020-04-10
Add pull_policy to the spec

### LINKED ISSUE #26: Define pull policy
Specification require image to be pulled then built if absent, and allow implementation ot offer options to tweak this behaviour. Users should be able to choose what's the best option for them.

Introduce `pull_policy` (`always|never|if_not_present|build`) so that users can define within compose file how to manage image pulls vs build.

hangyan:
Signed-off-by: Hang Yan <hangyan@alauda.io>

add `pull_policy` to the spec, 

Fixes #26

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -642,12 +642,24 @@ as `[<registry>/][<project>/]<image>[:<tag>|@<digest>]`.
     image: my_private.registry:5000/redis
 ```    
 
-If the image does not exist on the platform, Compose implementations MUST attempt to pull it. Compose
+If the image does not exist on the platform, Compose implementations MUST attempt to pull it based on the `pull_policy`. Compose
 implementations with build support MAY offer alternative options for the end user to control precedence of
-pull over building the image from source, however pulling the image MUST be the default behaviour.
+pull over building the image from source, however pulling the image MUST be the default behavior.
 
 `image` MAY be omitted from a Compose file as long as a `build` section is declared. Compose implementations
 without build support MUST fail when `image` is missing from the Compose file.
+
+
+### pull_policy
+
+`pull_policy` defines the decisions Compose implementations will make when it starts to pull images. Possible values are:
+
+* `always`: Compose implementations SHOULD always pull the image from the registry.
+* `never`: Compose implementations SHOULD NOT pull the image from a registry and SHOULD rely on the platform cached image. If there is no cached image, a failure MUST be reported.
+* `if_not_present`: Compose implementations SHOULD pull the image only if it's not available in the platform cache.This SHOULD be the default option for Compose implementations without build support.
+
+If `pull_policy` and `build` both presents, Compose implementations SHOULD build the image by default. Compose implementations MAY override this behavior in the toolchain.
+
 
 ### init
 
--- a/schema/config_schema_v3.9.json
+++ b/schema/config_schema_v3.9.json
@@ -162,6 +162,9 @@
         "healthcheck": {"$ref": "#/definitions/healthcheck"},
         "hostname": {"type": "string"},
         "image": {"type": "string"},
+        "pull_policy": {"type": "string", "enum": [
+          "always", "never", "if_not_present"
+        ]},
         "init": {"type": "boolean"},
         "ipc": {"type": "string"},
         "isolation": {"type": "string"},


spec.md
```diff
@@ -637,13 +637,25 @@ as `[<registry>/][<project>/]<image>[:<tag>|@<digest>]`.
...
 
 `image` MAY be omitted from a Compose file as long as a `build` section is declared. Compose implementations
 without build support MUST fail when `image` is missing from the Compose file.
 
+
+### pull_policy
+
+`pull_policy` define the decisions Compose implementations will make when it starts to pull images. Possible values are:
+
+* `always`: Compose implementations SHOULD always pull the image whether it's been pulled or not.
```
chris-crone, line 652:
```suggestion
* `always`: Compose implementations SHOULD pull the image from the registry ignoring other caches.
```

spec.md
```diff
@@ -637,13 +637,25 @@ as `[<registry>/][<project>/]<image>[:<tag>|@<digest>]`.
...
+* `never`: Compose implementations SHOULD NOT pull the image, instead it SHOULD use the image on the local.
```
chris-crone, line 653:
```suggestion
* `never`: Compose implementations SHOULD NOT pull the image from a registry and SHOULD rely on the platform cached image. If there is no cached image, a failure MUST be reported. 
```

spec.md
```diff
@@ -637,13 +637,25 @@ as `[<registry>/][<project>/]<image>[:<tag>|@<digest>]`.
...
+* `if_not_present`: Compose implementations SHOULD pull the image only if it's not found on the local.This SHOULD be the default option for Compose implementations without build support.
```
chris-crone, line 654:
```suggestion
* `if_not_present`: Compose implementations SHOULD pull the image from the registry only if it's not available in the platform cache. This SHOULD be the default option for Compose implementations without build support.
```

spec.md
```diff
@@ -637,13 +637,25 @@ as `[<registry>/][<project>/]<image>[:<tag>|@<digest>]`.
...
-pull over building the image from source, however pulling the image MUST be the default behaviour.
+pull over building the image from source, however pulling the image MUST be the default behavior.
 
 `image` MAY be omitted from a Compose file as long as a `build` section is declared. Compose implementations
 without build support MUST fail when `image` is missing from the Compose file.
 
+
+### pull_policy
+
+`pull_policy` define the decisions Compose implementations will make when it starts to pull images. Possible values are:
```
chris-crone, line 650:
```suggestion
`pull_policy` defines the decisions Compose implementations will make when it starts to pull images. Possible values are:
```

spec.md
```diff
@@ -637,13 +637,25 @@ as `[<registry>/][<project>/]<image>[:<tag>|@<digest>]`.
...
+
+* `always`: Compose implementations SHOULD always pull the image whether it's been pulled or not.
+* `never`: Compose implementations SHOULD NOT pull the image, instead it SHOULD use the image on the local.
+* `if_not_present`: Compose implementations SHOULD pull the image only if it's not found on the local.This SHOULD be the default option for Compose implementations without build support.
+
+`pull_policy` MAY be omitted from a Compose file as long as a `build` section is declared.
```
chris-crone, line 656: We say higher up that the default is `if_not_present` so we need to make this language clear for that case.

What about if both `pull_policy` and `build` are declared?

hangyan: Thanks for all these great advices(Sorry about my written english...). I think allow user to specify both `build` and `pull_policy` is  not necessary, simply report an error should be a good choice.

chris-crone: No problem!

> I think allow user to specify both build and pull_policy is not necessary, simply report an error should be a good choice.

I can see a case where you would want both build and pull policy set: if you're building your images locally to testing them and then want deploy the same Compose file to a remote endpoint. Ideally you wouldn't have to edit the Compose file between those steps to replace the build section with a pull policy.

An imagined UX for this could be:
```console
$ docker-compose up --build # locally build images and run
$ docker-compose push # push all built images to a registry
$ docker-compose up --context remote # run remotely using pull policy
```

The above assumes that by default we try to pull images (respecting the pull policy) and have an override flag to build. This logic could be reversed and we may want different logic for different implementations.

A straw man that allows implementations to decide could be:
> If `pull_policy` and `build` are omitted then implementations MUST pull the image using the default pull policy of `if_not_present`. If both `build` and `pull_policy` are set then implementations MAY build the image or MAY pull it respecting the pull policy. If only `build` is set then implementations MAY build the image or pull it using the default pull policy.

hangyan: Seems a good idea to use cmd flags. But in the spec, i'm afraid the aboving description can be confused for users and implementations. Since we can use cmd flags to override the default behavior, can we difine the default behavior as `build > pull_policy` in the spec, and make it clear that implements can provide cli flags or other ways to override this.

I'm not sure if i made my point clear.

chris-crone: I like the idea of having a SHOULD in the spec and explicitly saying that implementations can override this behaviour

hangyan: Updated

spec.md
```diff
@@ -642,13 +642,25 @@ as `[<registry>/][<project>/]<image>[:<tag>|@<digest>]`.
...
 
 `image` MAY be omitted from a Compose file as long as a `build` section is declared. Compose implementations
 without build support MUST fail when `image` is missing from the Compose file.
 
+
+### pull_policy
+
+`pull_policy` defines the decisions Compose implementations will make when it starts to pull images. Possible values are:
+
+* `always`: Compose implementations SHOULD always pull the image from the registry ignoring other caches.
```
Toasterson, line 657: Is there a reason for ignoring all caches? This is not necessary given the design of the OCI images. Even if the image has changed it may only have changed the last layer so the previous layers can be re-used from a local cache. According to the image spec that should be the default IIRC. Would be nice to clarify this. Not all implementations have to copy what is used in Kubernetes 1 to 1. Especially if a behavior can cause massive DDOS storms if used incorrectly.

hangyan: Personally i don't think this means ignore all caches

chris-crone: Coming from the `docker pull` perspective, I read "pull the image from the registry" to mean fetch the manifest for the tag from the registry and if it doesn't match the node cache then pull the layers that diff.

We may need to adjust the wording to make the intention more clear.

spec.md
```diff
@@ -642,13 +642,25 @@ as `[<registry>/][<project>/]<image>[:<tag>|@<digest>]`.
...
+* `never`: Compose implementations SHOULD NOT pull the image from a registry and SHOULD rely on the platform cached image. If there is no cached image, a failure MUST be reported.
+* `if_not_present`: Compose implementations SHOULD pull the image only if it's not available in the platform cache.This SHOULD be the default option for Compose implementations without build support.
+
+If `pull_policy` and `build` both presents, Compose implementations SHOULD build the image by default.Compose implementations MAY override this behavior in the tool chain.
```
chris-crone, line 661: Missing a space after a full stop.

```suggestion
If `pull_policy` and `build` both presents, Compose implementations SHOULD build the image by default. Compose implementations MAY override this behavior in the tool chain.
```

hangyan:
I see the json schema is not updated often? i guess it should be updated when new filed are added?

PR #60, opened 2020-04-11
Add regex format restriction for container_name

### LINKED ISSUE #34: Naming limitations for `container_name`
If I recall correctly, the `container_name` field is more restrictive than just _any_ UTF-8-encoded string. That should be specified in the spec.

hangyan:
Signed-off-by: Hang Yan <hangyan@alauda.io>

Add regex name restriction fort container_name feld.

Fixes #34

Seems this regex restrictions does not appear in the runtime spec, and it's reported by this issue 
https://github.com/moby/moby/issues/3138. Accounding the comments, this regex also apply for users and images, so i think it's fair enough to add this restirctions. As for kubernetes, it can easily use some transform methods to fit in this restriction.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -321,7 +321,7 @@ container_name: my-web-container
 Compose implementation MUST NOT scale a service beyond one container if the Compose file specifies a 
 `container_name`. Attempting to do so MUST result in an error.
 
-If represents, `container_name` SHOULD follow the regex fomrat of `[a-zA-Z0-9][a-zA-Z0-9_.-]+`
+If present, `container_name` SHOULD follow the regex format of `[a-zA-Z0-9][a-zA-Z0-9_.-]+`
 
 ### credential_spec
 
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -316,6 +316,8 @@ container_name: my-web-container
 Compose implementation MUST NOT scale a service beyond 1 container if the Compose file specifies a
 `container_name`. Attempting to do so MUST result in an error.
 
+If represents, `container_name` SHOULD follow the regex fomrat of `[a-zA-Z0-9][a-zA-Z0-9_.-]+`
```
ndeloof, line 319: typo

hangyan: fixed

spec.md
```diff
@@ -321,6 +321,8 @@ container_name: my-web-container
 Compose implementation MUST NOT scale a service beyond one container if the Compose file specifies a 
 `container_name`. Attempting to do so MUST result in an error.
 
+If represents, `container_name` SHOULD follow the regex fomrat of `[a-zA-Z0-9][a-zA-Z0-9_.-]+`
```
justincormack, line 324: typo here "fomrat" not "format"

SuperSandro2000, line 324:
```suggestion
If represents, `container_name` SHOULD follow the regex format of `[a-zA-Z0-9][a-zA-Z0-9_.-]+`
```

PR #63, opened 2020-04-21
Readd support for `device_cgroup_rules`

### LINKED ISSUE #62: Add support for `device_cgroup_rules`
The Docker container engine offers the `--device-cgroup-rule` option to add devices to allowed devices list. This is very helpful if hardware needs to be accessed from within the container to make sure only a minimal set of hardware is accessible (helps to avoid the use of `--privileged`).

In Docker Compose format 2.3 and later this has been part of the Docker Compose spec. I would like to readd this field to the latest version of the Docker Compose spec.

**Additional context**
This is a "Container runtime configuration", which in turn is part of "Deploy" area and as defined in the VISION.md document, this area is in scope of Compose.

The `device_cgroup_rules` as specified in Compose v2 reference:
https://docs.docker.com/compose/compose-file/compose-file-v2/#device_cgroup_rules

agners:
This readds support for `device_cgroup_rules` which was present in
compose file format 2.4.

Signed-off-by: Stefan Agner <stefan.agner@toradex.com>

Readd the `device_cgroup_rules` property which was present in Compose 2.3. It will allow to use Compose files which enable devices selectively via cgroup rules.

Fixes #62

### DIFF
--- a/schema/config_schema_v3.9.json
+++ b/schema/config_schema_v3.9.json
@@ -160,6 +160,7 @@
             }
           ]
         },
+        "device_cgroup_rules": {"$ref": "#/definitions/list_of_strings"},
         "devices": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
         "dns": {"$ref": "#/definitions/string_or_list"},
         "dns_search": {"$ref": "#/definitions/string_or_list"},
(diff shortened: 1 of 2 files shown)


agners:
@ndeloof added a description to `spec.md`.

agners:
@ndeloof I guess this is ready to be merged then?

ndeloof:
@agners can you please rebase to fix conflicts, then I'll merge this one

agners:
@ndeloof rebased!

PR #71, opened 2020-05-03
Add ipc: service mode to the spec

### LINKED ISSUE #70: Add support for `ipc: service` syntax
`docker` supports `--ipc` flag, which according to [docs](https://docs.docker.com/engine/reference/run/#ipc-settings---ipc):
> Shared memory segments are used to accelerate inter-process communication at memory speed, rather than through pipes or through the network stack. Shared memory is commonly used by databases and custom-built (typically C/OpenMPI, C++/using boost libraries) high performance applications for scientific computing and financial services industries. If these types of applications are broken into multiple containers, you might need to share the IPC mechanisms of the containers, using "shareable" mode for the main (i.e. “donor”) container, and "container:<donor-name-or-ID>" for other containers.

Assuming I have an application that makes use of shared memory, I can bring it up with `docker run --ipc=container:name ...`. All good here 👍 

However, if I'm using `docker-compose` - I am unable to get the same behaviour. This is because, according to [docs](https://docs.docker.com/compose/compose-file/#domainname-hostname-ipc-mac_address-privileged-read_only-shm_size-stdin_open-tty-user-working_dir), it's simply a pass-through option. This forces developers to co-host everything in the same container, which is difficult to work with and doesn't scale. 

It would be amazing if `ipc` option in `docker-compose` supported `service:[service name]` syntax, similar to [network_mode](https://docs.docker.com/compose/compose-file/#network_mode). For example:
```
version: '2.4'
services:
  myrtr:
    image: myipcrouter
  mysvc:
    build:
      context: .
      dockerfile: Dockerfile
    ipc: service:myrtr
```
This would make the container for the service (i.e, mysvc) have `--ipc=container:<id of myrtr>`.

**Additional context**
Original issue submitted here https://github.com/docker/compose/issues/7220, PR for implementation is available here https://github.com/docker/compose/pull/7417. @ndeloof suggested an issue/PR is submitted here as well.

EricHripko:
This PR aligns the `ipc` support for `service:[name]` syntax with existing functionality of `pid` and `network_mode`. The problem at hand is described in the linked issue (#70). The description of supported options was taken from [Docker docs](https://docs.docker.com/engine/reference/run/#ipc-settings---ipc).

In the words of @ndeloof in https://github.com/docker/compose/pull/7417:

> Sounds reasonable to me to align with other comparable attributes, but should first be discussed in https://github.com/compose-spec/compose-spec to ensure this get well set as supported compose syntax

Fixes #70.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1244,7 +1244,20 @@ expressed in the short form.
 
 ### ipc
 
-`ipc` configures the IPC isolation mode set by service container.
+`ipc` configures the IPC isolation mode set by service container. Available
+values are platform specific, but Compose specification defines specific values
+which MUST be implemented as described if supported:
+
+- `shareable` which gives the container own private IPC namespace, with a
+  possibility to share it with other containers.
+- `service:{name}` which makes the container join another (`shareable`)
+   container's IPC namespace.
+
+```yml
+    ipc: "shareable"
+    ipc: "service:[service name]"
+```
+
 
 ### mac_address
 


EricHripko:
Thank you for the approval @ndeloof 🙌 Based on the discussion in the linked issue, should I update the wording to include the following?

> Available values are platform specific, but Compose specification define specific values which MUST be implemented as described if supported.

(This is taken from `network_mode` attribute in the spec)

ndeloof:
I indeed think it would make sense to let implementation define custom values for this attribute, but enforce meaning for the ones we inherit from docker-compose.

EricHripko:
Cool 👍 I've reworded it to follow the same structure as `network_mode`. Feel free to take a look when you got a moment 🙂

justincormack:
This looks good to me. I think if an implementation supports ipc sharing this should be workable for them.

EricHripko:
Thanks for approving this 🙌 I've rebased on the latest `master` now too. Is this good to land?

EricHripko:
Hey @ndeloof @justincormack - just poking you to see if this can be landed 👀

EricHripko:
Wohoo - thank you for merging this 🙂

PR #72, opened 2020-05-03
Support condition: service_healthy

### LINKED ISSUE #68: Support condition: service_healthy from v2
We use Docker primarily for developing locally and running integration/system
tests on CI. This approach works brilliantly for us, as `docker-compose.yaml`
wholly describes the runtime environment needed to run an application. However,
we have a tough time dealing with the readiness problem. It's not enough to
just start the Compose services in the right order, as some infrastructure
takes time to become ready to serve the requests.

This negatively affects the experience with `docker-compose`, both when working
locally and, especially, when designing CI tests. This is because the 'naive
approach' of only using `depends_on` leads to a race condition by default. Such
tests will succeed if infrastructure comes up in time, and fail otherwise. 
Developers quickly learn to disregard flaky tests, at which point these are as
useful as having no tests.

**Additional context**
The general stance on this from Docker folks seems to summarised in 
[Startup order](https://docs.docker.com/compose/startup-order/) docs page:

> Your application needs to be resilient to these types of failures. To handle
> this, design your application to attempt to re-establish a connection to the
> database after a failure. If the application retries the connection, it can
> eventually connect to the database.

While the sentiment is definitely positive, I disagree with tightly coupling
readiness with resilience against network loss. An app can be resilient, but
choose to do sanity checks at startup. For example, a service could check if
a database is available at startup and, if it's down, exit out immediately.
This will cause the supervisor to restart the service and, eventually, begin
alarming to let SREs know that something is wrong. In my opinion, this approach
is just as valid as having the service remain in 'starting' state until the
database is available.

What is more, this seems to be a very deployment-oriented advice. If you use
`docker-compose` for working locally or CI, you probably just want the
containers to be ready and aren't too worried about resilience.

All of the workarounds suggest injecting some form of 'wait' script into the
command or entry-point. While this works for simpler setups, it doesn't really
scale well:

- __Duplication__ - dependencies are now listed in both `depends_on` and the
  wait script(s). Additionally, wait scripts now have to be (more or less)
  duplicated in all projects using `docker-compose`.
- __Leaky abstraction__ - developers now have to know how to check if a
  particular bit of infrastructure is ready (e.g., is it TCP? what's the port?
  is it HTTP? what URL should I check?). This becomes increasingly difficult in
  CI situations, where the database/cache/queue is pre-populated with data for
  reproducible tests (since just checking a port is no longer enough).
- __More dependencies__ - additional software is now required inside the app
  container. This degrades the overall UX from using Docker, as images are now
  larger and take longer to rebuild/download. In an anecdotal example, why
  should developers install several scripting interpreters just to check if
  dependencies are ready?
- __Complexity__ - each individual wait script needs to implement functionality
  that `healthcheck` already provides: initial delay, retrying and eventual
  timeout.

I believe that `condition: service_healthy` is a suitable and very elegant
solution to this issue. It would be great to carry it over from _version 2_ of
compose to the newer versions of the format.

It doesn't suffer from the same issues that the workarounds have:

- There is a __single source of truth__ when it comes to container dependencies
  (i.e., `depends_on` in the `docker-compose.yaml`).
- __Separation of concerns__ - infrastructure containers are
  responsible for performing health checks on themselves and indicating
  readiness for the main app container.
- __Works out-of-the-box__ - no additional software has to be installed in
  the app container in order to 'wait' for dependencies.
- __Simplicity__ - opting in for readiness is as easy as setting
  `condition: service_healthy` for your service.

EricHripko:
This PR imports `condition: service_healthy` functionality from Compose v2. The problem at hand was described and is being discussed in #68. Both the definition and the example were lifted from [Compose v2 docs](https://docs.docker.com/compose/compose-file/compose-file-v2/#depends_on).

Fixes #68.

### DIFF
--- a/schema/config_schema_v3.9.json
+++ b/schema/config_schema_v3.9.json
@@ -134,7 +134,28 @@
           },
           "additionalProperties": false
         },
-        "depends_on": {"$ref": "#/definitions/list_of_strings"},
+        "depends_on": {
+          "oneOf": [
+            {"$ref": "#/definitions/list_of_strings"},
+            {
+              "type": "object",
+              "additionalProperties": false,
+              "patternProperties": {
+                "^[a-zA-Z0-9._-]+$": {
+                  "type": "object",
+                  "additionalProperties": false,
+                  "properties": {
+                    "condition": {
+                      "type": "string",
+                      "enum": ["service_started", "service_healthy"]
+                    }
+                  },
+                  "required": ["condition"]
+                }
+              }
+            }
+          ]
+        },
         "devices": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
         "dns": {"$ref": "#/definitions/string_or_list"},
         "dns_search": {"$ref": "#/definitions/string_or_list"},
--- a/spec.md
+++ b/spec.md
@@ -369,8 +369,12 @@ configs:
 
 ### depends_on
 
-`depends_on` expresses a startup and shutdown dependencies between services, Service dependencies cause the following
-behaviors:
+`depends_on` expresses startup and shutdown dependencies between services.
+
+#### Short syntax
+
+The short syntax variant only specifies service names of the dependencies.
+Service dependencies cause the following behaviors:
 
 - Compose implementations MUST create services in dependency order. In the following
   example, `db` and `redis` are created before `web`.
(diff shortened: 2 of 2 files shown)


EricHripko:
@ndeloof I've addressed your feedback 🙂

EricHripko:
Hey @ndeloof! Thank you for approving my PR - really appreciate it 🙌  Is there anything else I could do to get this landed?

ndeloof:
rebased and merged

EricHripko:
Thank you for merging this 🙌 😄

ulgens:
@ndeloof @EricHripko What happened here? The branch seems not merged and deleted but the commits are in the main branch. Did someone apply an external patch instead of merging the PR?

ulgens:
Also, is this ever implemented in docker-compose?

ndeloof:
we don't use merge commits but rebase, so the commit to appear on master without an explicit merge commit

ulgens:
> we don't use merge commits but rebase

Rebase merge changes PR status to "merged" afaik. In the repo history there are multiple PRs which their latest status is "merged" but this one in particular shows as "closed", that's where I am lost.

EricHripko:
> Also, is this ever implemented in docker-compose?

Yes, this has been supported by Docker Compose implementation since [version 1.27.0](https://github.com/docker/compose/releases/tag/1.27.0) (although V2 potentially might have broken compatibility in places).

ulgens:
@EricHripko Thank you. Is this documented in the changelog? I read it three times but couldn't see anything related to condition keys.

EricHripko:
It's under the first point - _Merge 2.x and 3.x compose formats and align with COMPOSE_SPEC schema_. There's a bunch of other V2 functionality that became mainstream again with the arrival of Compose spec, which is why `condition: service_healthy` isn't called out explicitly.

PR #77, opened 2020-05-14
docs: typo top-level element configs

Schneereich:
**What this PR does / why we need it:**
According to https://docs.docker.com/compose/compose-file/#configs-configuration-reference the top-level element `configs:` is used in plural.


ndeloof:
Please sign-off your commits for contribution

Issue #12, opened 2020-01-10
Investigate v2 support


ndeloof:
**What is the problem you're trying to solve**
Docker compose file format version 2.x is still in use by many people
And only differ from v3 by a few removed attribute + addition of `deploy`

**Describe the solution you'd like**
spec could accept 2.x format. Maybe oculd also reconsider some of the removed attributes for inclusion in v3. excluded attribute would then be documented but flagged as Deprecated

SuperSandro2000:
I heavily rely on ``extends`` and the limits for mem and cpu so leaving 2.x out of the spec is a big bummer for me.

agners:
@ndeloof hm, `device_group_rules` is missing from the above list. Have you generated that list somehow? What was your base?

ndeloof:
I made this list from a diff between schema versions (updated with a few missing ones)

The main point in this issue is not to make an exhaustive list of missing attributes, but to debate if we want to adopt this approach to resurect them. On agreement we could then actually introduce them in the spec with fixes comparable to #63

`extends` being a special case, we probably will have to discuss this legacy feature separately

agners:
@ndeloof thanks for the list update!

I guess there are two steps, discussing reintroduction of v2 options, and actually introduce them. Is your idea to introduce them to the spec in one go or just have a single issue (this one) to discuss which ones to introduce?

There are actually quite some options. I more and more think it will be hard to do introduce/discuss them in union likely will take a long time to sort out things out and come to an agreement.

How about we group them a bit, and create separate issues for each group?

Resource constraints:

* `blkio_config`
* `cpu_count` :
* `cpu_percent`
* `cpu_shares`
* `cpu_period`
* `cpu_quota`
* `cpu_rt_period`
* `cpu_rt_runtime`
* `cpus` :
* `cpuset`
* `mem_limit`
* `memswap_limit`
* `mem_swappiness`
* `mem_reservation`
* `oom_kill_disable`
* `oom_score_adj`
* `pids_limit`

Devices

* `device_cgroup_rules`
* `devices`

Network
* `dns_opts`

Volume/Storage

* `storage_opt`
* `volume_driver`
* `volumes_from`

Others
* `group_add`
* `platform`
* `runtime`
* `scale` (replaced by `replicas`?)
* `extends` (*)

ndeloof:
This is really about discussing reintroduction of v2 options. List was most to illustrate the diff. Actual execution will be driven by our decision.

I can see 3 categories:
[...]

PR #78, opened 2020-05-15
(re)introduce some of the compose v2 attributes

ndeloof:

Reintroduce attributes from compose 2.x file format:
- build.extra_hosts
- build.isolation
- blkio_config
- cpu_count
- cpu_percent
- cpu_shares
- cpu_period
- cpu_quota
- cpu_rt_runtime
- cpu_rt_period
- cpus
- cpuset
- dns_opt
- group_add
- oom_kill_disable
- oom_score_adj
- pid_limit
- platform
- runtime
- scale

https://github.com/compose-spec/compose-spec/issues/12

### DIFF
--- a/schema/config_schema_v3.9.json
+++ b/schema/config_schema_v3.9.json
@@ -91,12 +91,41 @@
                 "cache_from": {"$ref": "#/definitions/list_of_strings"},
                 "network": {"type": "string"},
                 "target": {"type": "string"},
-                "shm_size": {"type": ["integer", "string"]}
+                "shm_size": {"type": ["integer", "string"]},
+                "extra_hosts": {"$ref": "#/definitions/list_or_dict"},
+                "isolation": {"type": "string"}
               },
               "additionalProperties": false,
               "patternProperties": {"^x-": {}}
             }
           ]
+        },
+        "blkio_config": {
+          "type": "object",
+          "properties": {
+            "device_read_bps": {
+              "type": "array",
+              "items": {"$ref": "#/definitions/blkio_limit"}
+            },
+            "device_read_iops": {
+              "type": "array",
+              "items": {"$ref": "#/definitions/blkio_limit"}
+            },
+            "device_write_bps": {
+              "type": "array",
+              "items": {"$ref": "#/definitions/blkio_limit"}
+            },
+            "device_write_iops": {
+              "type": "array",
+              "items": {"$ref": "#/definitions/blkio_limit"}
+            },
+            "weight": {"type": "integer"},
+            "weight_device": {
+              "type": "array",
+              "items": {"$ref": "#/definitions/blkio_weight"}
+            }
+          },
+          "additionalProperties": false
         },
         "cap_add": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
         "cap_drop": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
@@ -128,6 +157,15 @@
           }
         },
         "container_name": {"type": "string"},
+        "cpu_count": {"type": "integer", "minimum": 0},
+        "cpu_percent": {"type": "integer", "minimum": 0, "maximum": 100},
+        "cpu_shares": {"type": ["number", "string"]},
+        "cpu_quota": {"type": ["number", "string"]},
+        "cpu_period": {"type": ["number", "string"]},
+        "cpu_rt_period": {"type": ["number", "string"]},
+        "cpu_rt_runtime": {"type": ["number", "string"]},
+        "cpus": {"type": "number", "minimum": 0},
+        "cpuset": {"type": "string"},
         "credential_spec": {
           "type": "object",
           "properties": {
@@ -163,6 +201,13 @@
         "device_cgroup_rules": {"$ref": "#/definitions/list_of_strings"},
         "devices": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
         "dns": {"$ref": "#/definitions/string_or_list"},
+        "dns_opt": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "uniqueItems": true
+        },
         "dns_search": {"$ref": "#/definitions/string_or_list"},
         "domainname": {"type": "string"},
         "entrypoint": {
@@ -185,18 +230,21 @@
 
         "external_links": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
         "extra_hosts": {"$ref": "#/definitions/list_or_dict"},
+        "group_add": {
+          "type": "array",
+          "items": {
+            "type": ["string", "number"]
+          },
+          "uniqueItems": true
+        },
         "healthcheck": {"$ref": "#/definitions/healthcheck"},
         "hostname": {"type": "string"},
         "image": {"type": "string"},
-        "pull_policy": {"type": "string", "enum": [
-          "always", "never", "if_not_present"
-        ]},
         "init": {"type": "boolean"},
         "ipc": {"type": "string"},
         "isolation": {"type": "string"},
         "labels": {"$ref": "#/definitions/list_or_dict"},
         "links": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
-
         "logging": {
             "type": "object",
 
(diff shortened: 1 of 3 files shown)


spec.md
```diff
@@ -704,25 +841,13 @@ as `[<registry>/][<project>/]<image>[:<tag>|@<digest>]`.
     image: my_private.registry:5000/redis
 ```    
 
-If the image does not exist on the platform, Compose implementations MUST attempt to pull it based on the `pull_policy`. Compose
+If the image does not exist on the platform, Compose implementations MUST attempt to pull it based on the `pull_olicy`. Compose
```
EricHripko, line 844: `pull_olicy` is a typo

spec.md
```diff
@@ -484,6 +592,16 @@ dns:
   - 9.9.9.9
 ```
 
+### dns_opt
+
+`don_opt` list custom DNS options to be passed to the container’s DNS resolver (`/etc/resolv.conf` file on Linux).
```
EricHripko, line 597: `don_opt` typo

spec.md
ndeloof: nice catch, thanks

spec.md
```diff
@@ -188,6 +188,114 @@ not implemented the Deploy section SHOULD be ignored and the Compose file MUST s
...
+
+`cpu_rt_runtime` configures CPU allocation parameters for platform with support for realtime scheduler. Can be either 
+an integer value using microseconds as unit or a [duration](#specifying-durations).
+
+```yml
+ cpu_rt_runtime: '400ms'
+ cpu_rt_runtime: 95000`
+```
+
+### cpu_rt_period🔗
```
EricHripko, line 250: Looks like an emoji slipped in here 🙂

spec.md
```diff
@@ -188,6 +188,114 @@ not implemented the Deploy section SHOULD be ignored and the Compose file MUST s
...
+_DEPRECATED: use [deploy.reservations.cpus](deploy.md#cpus)_
+
+`cpus` define the number of (potentially virtual) CPUs to allocate to service containers. This is a fractional number. 
+`0.000` means no limit.
+
+### cpuset
+
+`cpuset` defines the explicit CPUs in which to allow execution. Can be a range `0-3` or a list `0,1`
+
+#### device_read_bps, device_write_bps
```
EricHripko, line 270: Should these be under `blkio_config` header?

schema/config_schema_v3.9.json
```diff
@@ -242,8 +288,11 @@
             }
           ]
         },
+        "oom_kill_disable": {"type": "boolean"},
+        "oom_score_adj": {"type": "integer", "minimum": -1000, "maximum": 1000},
         "pid": {"type": ["string", "null"]},
-
+        "pids_limit": {"type": ["number", "string"]},
```
thaJeztah, line 294: Can we put this under `resources`?

schema/config_schema_v3.9.json
```diff
@@ -91,13 +91,42 @@
...
+              "type": "array",
+              "items": {"$ref": "#/definitions/blkio_limit"}
+            },
+            "device_write_bps": {
+              "type": "array",
+              "items": {"$ref": "#/definitions/blkio_limit"}
+            },
+            "device_write_iops": {
+              "type": "array",
+              "items": {"$ref": "#/definitions/blkio_limit"}
```
thaJeztah, line 120: We should consider re-designing this, e.g. to group these options _per device_

```yaml
devices:
    - device: <device>
      iops: <x>
      bps: <x>
```

schema/config_schema_v3.9.json
ndeloof: We are discussing re-introducing support for V2 schema attributes, this doesn't prevent another discussion about deprecating some of them and introduce a better way to define equivalent feature

schema/config_schema_v3.9.json
ndeloof: I also think so.
We also should define an `oci` (or `linux` or any comparable name) for cgroup related stuff and raw linux kernel features exposed by the OCI runtime

schema/config_schema_v3.9.json
```diff
@@ -128,6 +157,15 @@
...
         "container_name": {"type": "string"},
+        "cpu_count": {"type": "integer", "minimum": 0},
+        "cpu_percent": {"type": "integer", "minimum": 0, "maximum": 100},
+        "cpu_shares": {"type": ["number", "string"]},
+        "cpu_quota": {"type": ["number", "string"]},
+        "cpu_period": {"type": ["number", "string"]},
+        "cpu_rt_period": {"type": ["number", "string"]},
+        "cpu_rt_runtime": {"type": ["number", "string"]},
+        "cpus": {"type": "number", "minimum": 0},
+        "cpuset": {"type": "string"},
```
thaJeztah, line 168: There's an overlap between some of these and `resource.cpus`.

Some of them are really hard to use, and must be specified "together" (`cpu-period` and `cpu-quota`, but also `cpuset-cpus` and `cpuset-mem`). Ideally we would have a better design for that to make such things logically grouped, so that they will be set together https://docs.docker.com/engine/reference/run/#cpu-period-constraint

schema/config_schema_v3.9.json
```diff
@@ -185,18 +230,21 @@
 
         "external_links": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
         "extra_hosts": {"$ref": "#/definitions/list_or_dict"},
+        "group_add": {
+          "type": "array",
+          "items": {
+            "type": ["string", "number"]
+          },
+          "uniqueItems": true
+        },
```
thaJeztah, line 239: Need to review how these (should) interact with `user`; `user` can have `user:group`, which sets the primary group, and this sets auxiliary groups. For swarm services we renamed this to `groups`

```
--group list  Set one or more supplementary user groups for the container
```

spec.md
```diff
@@ -188,6 +188,114 @@ not implemented the Deploy section SHOULD be ignored and the Compose file MUST s
...
+  foo:
+    image: busybox
+    blkio_config:
+       weight: 300
+       weight_device:
+         - path: /dev/sda
+           weight: 400
+       device_read_bps:
+         - path: /dev/sdb
+           rate: '12mb'
```
thaJeztah, line 206: See my other comment; these settings are _per device_; ideally we would _flip_ the configuration to be grouped per device. These configs are also non-portable (devices will be different per environment, which makes it not very suitable for a compose-file that is to be shared between developers / environments)

spec.md
```diff
@@ -925,11 +1050,44 @@ networks:
         - subnet: "2001:3984:3989::/64"
 ```
 
+### oom_kill_disable
+ 
+If `oom_kill_disable` is set Compose implementation MUST configure the platform so it won't kill the container in case 
+of memory starvation.
```
thaJeztah, line 1056: There's been some discussion about this option overall; it's _really_ dangerous (process can never be killed, which means the host will be killed before this container). I'd be very cautious bringing this back.

`oom_score_adj` would already achieve this in 99% of the situations. (`-1000` is "unkillable")

spec.md
```diff
@@ -991,6 +1149,18 @@ ports:
...
 ```
 
+
+### pull_policy
+
+`pull_policy` defines the decisions Compose implementations will make when it starts to pull images. Possible values are:
+
+* `always`: Compose implementations SHOULD always pull the image from the registry.
+* `never`: Compose implementations SHOULD NOT pull the image from a registry and SHOULD rely on the platform cached image. If there is no cached image, a failure MUST be reported.
+* `if_not_present`: Compose implementations SHOULD pull the image only if it's not available in the platform cache.This SHOULD be the default option for Compose implementations without build support.
```
thaJeztah, line 1159: This is the equivalent of `--pull=missing` (https://github.com/docker/cli/pull/1498)? Was this how it was named in the v2 format, or renamed from what we had previously?

spec.md
ndeloof: again, we bring it back to the spec because it existed in v2 and we want to give implementation a chance to parse any compose file and manage such attribute the way they want (can just ignore with a warning). We can discuss about deprecating any dangerous feature in another issue

spec.md
ndeloof: I guess it was isnpired by Kubernetes `IfNotPresent` pull policy

ndeloof: note: this one was introduced previously, only moved to use alphabetic ordering on attributes

PR #86, opened 2020-06-19
Specify the behaviour of extends field

### LINKED ISSUE #67: Consider support for compose v2 `extends`
`extends` was supported in compose file format v2 and removed in v3. Has since be requested by many users, see https://github.com/moby/moby/issues/31101

Re-introduce this feature, with a full description so we don't suffer implementation glitches between docker-compose's _de facto_ python implementation and compose-go

**Additional context**
see https://github.com/docker/compose/issues/4315 for debates about the validity of this feature and it's drawbacks. Also consider option to use of compose file overrides and yaml anchors to achieve comparable results.

EricHripko:
This PR specifies the behaviour of `extends` field. This should complete the merger of `v2`/`v3` specs and enable us to release the first iteration of Compose spec.

Fixes #67.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -747,7 +747,7 @@ expose:
 ### extends
 
 Extend another service, in the current file or another, optionally overriding configuration. You can use
-`extends` on any service together with other configuration keys. The `extends` value must be a mapping
+`extends` on any service together with other configuration keys. The `extends` value MUST be a mapping
 defined with a required `service` and an optional `file` key.
 
 ```yaml
--- a/spec.md
+++ b/spec.md
@@ -783,10 +783,9 @@ Compose implementations MUST return an error in all of these cases.
 - File path, which can be either:
   - Relative path. This path is considered as relative to the location of the main Compose
     file.
-  - Absolute path. `~/` in path is expanded to the home directory of current user. `~user/`
-    is expanded to the home directory of `user` user.
+  - Absolute path.
 
-Service denoted by `service` must be present in the identified referenced Compose file.
+Service denoted by `service` MUST be present in the identified referenced Compose file.
 Compose implementations MUST return an error if:
 
 - Service denoted by `service` was not found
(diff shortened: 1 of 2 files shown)


EricHripko:
Also cc @kohidave since he's on the Maintainers PR.

EricHripko:
Hey folks! This is now ready for review - I've put this together by scanning through https://github.com/docker/compose code and experimenting with `1.25.5` version of `docker-compose`. All the behaviours were confirmed to be working with the CLI by invoking `docker-compose config` and inspecting the merged configuration. The only exception to that is `deploy` mapping, which is currently specific to v3 (and thus incompatible with `extends` from v2).

spec.md
```diff
@@ -744,6 +744,202 @@ expose:
   - "8000"
 ```
 
+### extends
+
+Extend another service, in the current file or another, optionally overriding configuration. You can use
+`extends` on any service together with other configuration keys. The `extends` value must be a mapping
```
ndeloof, line 750: nit: use uppercase `MUST` as other sections

spec.md
```diff
@@ -744,6 +744,202 @@ expose:
...
+
+Compose implementations MUST return an error in all of these cases.
+
+#### Finding referenced service
+
+`file` value can be:
+
+- Not present.
+  This indicates that another service within the same Compose file is being referenced.
+- File path, which can be either:
```
ndeloof, line 783: In other section we recommend not using relative path to enforce portability. Maybe we should introduce a dedicated section and just have references where relevant. I didn't know about `~user`, this sounds a bit hack-ish and highly unportable. I'm not sure we should have this being part of the spec (can be a docker-compose implementation detail)

spec.md
EricHripko: Done!

spec.md
EricHripko: Removed the user path expansion bit (~ and ~user). I'm open to creating a separate section for relative paths 🙂

EricHripko:
I've applied the feedback action discussed at the Community Meeting yesterday. @justincormack @hangyan @ndeloof @kohidave could you take a look? 👀

ndeloof:
manually rebased and merged

PR #89, opened 2020-07-10
Add more v2.x and v3.x fields

aiordache:
Align schema with the unified schema from docker-compose.

Relates to https://github.com/docker/compose/issues/7201

Signed-off-by: aiordache <anca.iordache@docker.com>

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -630,7 +630,8 @@
               "patternProperties": {"^.+$": {"type": "string"}}
             }
           },
-          "additionalProperties": false
+          "additionalProperties": false,
+          "patternProperties": {"^x-": {}}
         },
         "external": {
           "type": ["boolean", "object"],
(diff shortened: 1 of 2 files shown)


schema/compose-spec.json
```diff
@@ -606,15 +611,26 @@
...
+            },
+            "options": {
+              "type": "object",
+              "additionalProperties": false,
+              "patternProperties": {"^.+$": {"type": "string"}}
             }
           },
-          "additionalProperties": false,
-          "patternProperties": {"^x-": {}}
+          "additionalProperties": false
```
EricHripko, line 633: Shouldn't we keep `patternProperties` here or is the removal intentional? This means that no custom/experimental fields can be attached to `ipam` config.

aiordache: My mistake, I'll add it. Thanks for noticing it!

ndeloof: nice catch!

ndeloof:
@aiordache pushed documentation for missing fields. Please do a quick check to confirm I didn't forgot one/made a bad copy/paste from compose v2 docs :)

PR #96, opened 2020-08-18
Add `enable_ipv6` property to networks and update `cpus` type

aiordache:
Added missing (documented) field to the schema and updated the `cpus` properties.


chris-crone:
SGTM

We should add this to compose-go as well if it isn't already there.

EricHripko (approved):
Looks great - thank you ✅

ndeloof (approved):

Issue #97, opened 2020-08-25
Proposal: profiles


ndeloof:
**What is the problem you're trying to solve**
Many users define in a compose file some additional container that they use during development as "one-off" tasks using `docker-compose run` command, but are not part of their compose application. As one can't disable a service in a compose file, they have to maintain a separated compose file, and then use explicit `--file` option.

It's also a common practice to have separated override compose file to address different environments, typically to configure some services with debug port exposed during development.

This proposal is about making those usage patterns more canonical by providing the adequate descriptive apporach to select a sub/superset of containers in a compose file, and apply some override files.

**Describe the solution you'd like**
Compose specification introduce a new concept as *profile* that a Compose implemenation can use to select the services to be ran when loading a Compose model.

1. Introduce a new `profiles` attribute at services level as a string list. If not set, it will default to `profiles: ["default"]`. In regards to https://github.com/compose-spec/compose-spec/issues/87 this is equivalent to `auto_up: true`

Compose implementations whith profile support MUST ignore services which have a profile attribute but don't match the current profile. Services without a profile attribute always match.

Selecting active profile(s) can be implemented by any mechanism, typically flags or environment variable `DOCKER_COMPOSE_PROFILE`, but this is an implementation detail not to be defined y the spec.

Running `compose up` with "debug" profile enabled will run: 
- services with no `profiles` attribute
- services which `profiles` list includes "debug" in the `profiles` attribute. 

2. Clarify by the specification the selection of an override compose file to be merged with the main one. With profile support, a Compose implementation will check for the presence of a `compose-[profile].yaml` file aside the main compose file, and automatically merge it with the main model. Adopting this pattern by the spec would actually make a common practice a canonical one - docker-compose already does this by loading a `docker-compose-overrides.yaml` file if present.

**Additional context**
This proposal is a follow-up for https://github.com/docker/compose/issues/7539#issuecomment-645361324 and address https://github.com/compose-spec/compose-spec/issues/87

rumpl:
I'm making a connection with Spring profiles, do we want to accept multiple profiles when running an application, eg. `compose run --profile debug --profile elk` for example?

rumpl:
If I understand correctly, if I run 

```
compose --profile test up
```
it will load and merge the `docker-compose-test.yaml` file. But if I run 

```
compose --profile default up
```
then it will load the `docker-compose-overrides.yaml` one, right?

If this is the case then we would need to make this very clear in the spec/documentation and maybe event while running the app.

[...]

PR #98, opened 2020-09-02
Introduce `profiles`

ndeloof:
Introduce `profiles` in compose schema and specification.
This is based on https://github.com/compose-spec/compose-spec/issues/97 discussion, but does not cover the `override` file proposal, which could be proposed in a follow-up PR.

see https://github.com/compose-spec/compose-spec/issues/97

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -167,6 +167,43 @@ merged are hosted in other folders.
 As some Compose file elements can both be expressed as single strings or complex objects, merges MUST apply to
 the expanded form.
 
+### Profiles
+
+Profiles allow to adjust the Compose application model for various usages and environments. A Compose
+implementation SHOULD allow user to define a set of active profiles. The exact mechanism is implementation
+specific and MAY include command line flags, environment variables, etc.
+
+All top-level elements support a `profiles` attribute to define a list of named profiles. Elements without 
+a `profiles` attribute set MUST always be enabled. The enclosing component MUST be ignored by compose
+implementation when none of the listed profiles match the active ones, until the target component is
+explicitly targeted by a command.
+
+References to other services (by `links`, `extends` or shared resource syntax `service:xxx`) MUST not
+automatically enable a component that would otherwise have been ignored by active profiles.
+
+#### illustration sample
+
+```yaml
+services:
+  foo:
+     image: foo
+  bar:
+     image: bar
+     profiles: ["test"]
+  zot:
+     image: zot
+     profiles: ["debug"]
+     depends_on: ["bar"]
+```
+
+- Compose application model parsed with no profile enabled only contains the `foo` service.
+- If profile `test` is enabled, model contains both the `foo` and `bar` service.
+- If profile `debug` is enabled, model contains both the `foo` and `zot` service, but not `bar` 
+  and as such the model is invalid regarding the `depends_on` constraint.
+- If Compose implementation is executed with `bar` as explicit service to run, it will be active 
+  whenever `test` profile is not enabled.
+
+
 ## Version top-level element
 
 Top-level `version` property is defined by the specification for backward compatibility but is only informative. 
@@ -1458,6 +1495,10 @@ ports:
 
 `privileged` configures the service container to run with elevated privileges. Support and actual impacts are platform-specific.
 
+### profiles
+
+`profiles` defines a list of named profile for service to be enabled. When not set, service is always enabled.
+
 ### pull_policy
 
 `pull_policy` defines the decisions Compose implementations will make when it starts to pull images. Possible values are:
@@ -1925,6 +1966,10 @@ labels:
 
 Compose implementations MUST set `com.docker.compose.project` and `com.docker.compose.network` labels.
 
+### profiles
+
+`profiles` defines a list of named profile for network to be enabled. When not set, network is always enabled
+
 ### external
 
 If set to `true`, `external` specifies that this network’s lifecycle is maintained outside of that of the application.
@@ -2098,6 +2143,11 @@ volumes:
       name: ${DATABASE_VOLUME}
 ```
 
+### profiles
+
+`profiles` defines a list of named profile for volume to be enabled. When not set, volume is always enabled.
+
+
 ## Configs top-level element
 
 Configs allow services to adapt their behaviour without the need to rebuild a Docker image. Configs are comparable to Volumes from a service point of view as they are mounted into service's containers filesystem. The actual implementation detail to get configuration provided by the platform can be set from the Configuration definition.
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -167,6 +167,43 @@ merged are hosted in other folders.
...
+services:
+  foo:
+     image: foo
+  bar:
+     image: bar
+     profiles: ["test"]
+  zot:
+     image: zot
+     profiles: ["debug"]
+     depends_on: ["bar"]
```
rumpl, line 196:
```suggestion
services: 
  foo: 
    image: foo
  bar: 
    image: bar
    profiles: 
      - test
  zot: 
    depends_on: 
      - bar
    image: zot
    profiles: 
      - debug
```

spec.md
```diff
@@ -167,6 +167,43 @@ merged are hosted in other folders.
...
+```
+
+- Compose application model parsed with no profile enabled only contains the `foo` service.
+- If profile `test` is enabled, model contains both the `foo` and `bar` service.
+- If profile `debug` is enabled, model contains both the `foo` and `zot` service, but not `bar`
```
rumpl, line 201:
```suggestion
- If profile `debug` is enabled, model contains both `foo` and `zot` services, but not `bar` 
```

spec.md
```diff
@@ -167,6 +167,43 @@ merged are hosted in other folders.
...
+Profiles allow to adjust the Compose application model for various usages and environments. A Compose
+implementation SHOULD allow user to define a set of active profiles. The exact mechanism is implementation
+specific and MAY include command line flags, environment variables, etc.
+
+All top-level elements support a `profiles` attribute to define a list of named profiles. Elements without 
+a `profiles` attribute set MUST always be enabled. The enclosing component MUST be ignored by compose
+implementation when none of the listed profiles match the active ones, until the target component is
+explicitly targeted by a command.
+
+References to other services (by `links`, `extends` or shared resource syntax `service:xxx`) MUST not
```
acran, line 181: See https://github.com/compose-spec/compose-spec/issues/97#issuecomment-685712571

acran:
With the follow-up PR for `override` files in mind: should we maybe narrow down the valid profile names? These should exclude all invalid characters for filesystem paths and maybe also `/` and `\`?

Also having a profile name start with '-' could make parsing command lines hard:
~~~sh
# start profile named "--build"
compose up --profile --build
~~~

ndeloof:
About restriction on profile name, I suggest we re-use work done on https://github.com/compose-spec/compose-spec/pull/60 and define `[a-zA-Z0-9][a-zA-Z0-9_.-]+` as a global pattern for "names".

spec.md
```diff
@@ -167,6 +167,43 @@ merged are hosted in other folders.
...
+automatically enable a component that would otherwise have been ignored by active profiles.
+
+#### illustration sample
```
EricHripko, line 184:
```suggestion
#### Illustration sample
```
Since other headings uppercase first letter

spec.md
```diff
@@ -2098,6 +2143,11 @@ volumes:
       name: ${DATABASE_VOLUME}
 ```
 
+### profiles
+
+`profiles` defines a list of named profile for volume to be enabled. When not set, volume is always enabled.
```
EricHripko, line 2148:
```suggestion
`profiles` defines a list of named profiles for volume to be enabled under. When not set, volume is always enabled.
```

spec.md
```diff
@@ -1925,6 +1966,10 @@ labels:
 
 Compose implementations MUST set `com.docker.compose.project` and `com.docker.compose.network` labels.
 
+### profiles
+
+`profiles` defines a list of named profile for network to be enabled. When not set, network is always enabled
```
EricHripko, line 1971:
```suggestion
`profiles` defines a list of named profiles for network to be enabled under. When not set, network is always enabled
```

spec.md
```diff
@@ -1458,6 +1495,10 @@ ports:
 
 `privileged` configures the service container to run with elevated privileges. Support and actual impacts are platform-specific.
 
+### profiles
+
+`profiles` defines a list of named profile for service to be enabled. When not set, service is always enabled.
```
EricHripko, line 1500:
```suggestion
`profiles` defines a list of named profiles for service to be enabled under. When not set, service is always enabled.
```

spec.md
```diff
@@ -167,6 +167,43 @@ merged are hosted in other folders.
...
+### Profiles
+
+Profiles allow to adjust the Compose application model for various usages and environments. A Compose
+implementation SHOULD allow user to define a set of active profiles. The exact mechanism is implementation
+specific and MAY include command line flags, environment variables, etc.
+
+All top-level elements support a `profiles` attribute to define a list of named profiles. Elements without 
+a `profiles` attribute set MUST always be enabled. The enclosing component MUST be ignored by compose
+implementation when none of the listed profiles match the active ones, until the target component is
+explicitly targeted by a command.
```
EricHripko, line 179: It feels a bit strange that users can get (what is essentially) a one-off profile with always-on services + just the target service. This might get rather confusing since we're ignoring `profiles` attribute on the target service, but enforcing it everywhere else.

Is it worth to fomalise this a bit more by saying that `compose run/up TARGET` will activate the `profiles` of the target service? I appreciate that this probably makes the lives of Compose implementors harder (especially considering the follow-up profile-specific files proposal), but this functionality in isolation seems more confusing than beneficial.

EricHripko (commented):
Looks like a great start and a lovely potential addition to the spec 🙌 I got one chunky point of discussion inline and a few minor suggestions 💬

EricHripko:
@ndeloof - have you had a chance to break this out into a smaller initial profile proposal?

ndeloof:
@EricHripko have been pretty busy, feel free to take ownership on this topic if you can

acran:
@ndeloof I'd gladly support here, what can I do to advance this PR?

AFAIK the next step was to reduce the scope of this PR to a minimum and add further use cases with additional PRs? But for me this PR already is quite limited in scope, so what should be cut for the first step? I would only think of limiting the profiles key word to the `service` model.

If that would be okay, I could create a new PR with this and incorporate the pending feedback from here (since I can not take over this PR??).

ndeloof:
@acran  yes indeed, we agreed on reducing scope to `profile` attribute introduction with minimal "magic" so we cover some basic use-cases and can have further debates on the more advanced ones.
Feel free to create another PR and cherry-pick any relevant parts from this one

acran:
I created a new PR with hopefully all the feedback incorporated in #110. Please have a look (=

EricHripko:
This is awesome - thanks for picking it up @acran 🙂

spec.md
muxator: The term "shared resource syntax `service:xxx`" is not defined anywhere in the spec document.

Where can a newcomer learn what is it? It would be good if there was at least a hyperlink here.

acran: This is referring to system resources shareable between services or with the host, where you can specify the resource with the `service:[service name]` syntax, see [`network_mode`](https://github.com/compose-spec/compose-spec/blob/9015d1b3e0cd0504e38b9d22f8845068c7f22584/spec.md#network_mode) and [`ipc`](https://github.com/compose-spec/compose-spec/blob/9015d1b3e0cd0504e38b9d22f8845068c7f22584/spec.md#ipc) in the spec.

muxator: Thanks for the clarification. If the document included these links to the spec that would be really helpful for other people, too. Do you think it can be added?

muxator: Now I see that this snippet was already merged in https://github.com/compose-spec/compose-spec/pull/110/files#diff-bc6661da34ecae62fbe724bb93fd69b91a7f81143f2683a81163231de7e3b545R182.

My comment should really be placed there. The gist of it stands: could a link [network_mode](https://github.com/compose-spec/compose-spec/blob/9015d1b3e0cd0504e38b9d22f8845068c7f22584/spec.md#network_mode) and [ipc](https://github.com/compose-spec/compose-spec/blob/9015d1b3e0cd0504e38b9d22f8845068c7f22584/spec.md#ipc) be added in the document?

ndeloof:
closing as obsolete

PR #100, opened 2020-09-05
Initial specification for accelerators

### LINKED ISSUE #74: Need support for non-CPU accelerators in docker-compose
The docker-compose specification does not support mixing GPU and CPU containers on the same machine. The runtime: tag that allowed us to do this was removed in 3.x. However, I believe that this is reflective of a larger problem that I'd like to see addressed. Devices will have multiple "accelerators" in addition to CPUs such as Nvidia GPU, AMD GPU, ARM GPU (pi), Google TPU, Intel TPU,... etc etc. Consequently, we need a clear way to request that a container gain access to a particular accelerator and that the orchestrator respects this when scheduling the container. Ideally orchestrators like Kube should also be able to play here.

A run time constraint that describes the kind of accelerator needed will be ideal and the plumbing needed for the container to map the accelerator specified should be set up when the orchestrator schedules the container onto the machine hosting the specified accelerator. 

**Additional context**
Please see Github Issue: [Nvidia docker-compose](https://github.com/docker/compose/issues/6691)

EricHripko:
This PR adds a section to the spec enabling Compose to make use of the graphics/AI accelerators. Most common use case is nVidia CUDA-enabled GPUs, but (as mentioned in the issue) there are some other devices like TPUs (for generic accelerated machine learning) and VPUs (for accelerated machine learning in computer vision). The spec closely mirrors `DeviceRequest` type with a single simplification of having `capabilities` as a flat list of strings (as also seen in https://github.com/docker/compose/issues/6691#issuecomment-670700674). 

This is pretty much essential for developers that make use of containers for machine learning and other AI tasks. Without accelerators, developers are stuck using CPU for these tasks - which is considerably slower.

**Potential mapping:**
- `docker-compose` can make use `DeviceRequest` to reserve the right devices
- For Kubernetes:
  - `nvidia` driver + `gpu` capability can get translated to `limits: nvidia.com/gpu: 1`
  - `amd` driver + `gpu` capability can get translated to `limits: amd.com/gpu: 1`
  - `gcloud` driver + `tpu` capability can get translated to `cloud-tpus.google.com/v2: 8`
- On ECS, `nvidia` driver + `gpu` capability can be translated to:
```json
        "resourceRequirements" : [
            {
               "type" : "GPU", 
               "value" : "2"
            }
        ],
```
- On ACI, `nvidia` driver + `gpu` capability can be translated to below. `sku` can either be derived from `count` or specified as an `x-` field
```
          gpu:
            count: 1
            sku: K80
```

**Shoutout to:**
- @Lucidiot for exposing `DeviceRequest` API in `docker-py`: https://github.com/docker/docker-py/pull/2471
- @bkakilli for putting together a sample implementation of `DeviceRequest` API in `docker-compose`: https://github.com/docker/compose/issues/6691#issuecomment-670700674
- @vk1z for submitting the issue and providing maintainers with valuable insights around accelerators and containers
- @justincormack, @chris-crone and @ndeloof for helping guide the discussions and making sure this issue doesn't slip through the cracks 

Fixes #74.

### DIFF
--- a/deploy.md
+++ b/deploy.md
@@ -201,7 +201,7 @@ deploy:
 
 ##### count
 
-If `count` is set to `-1` or not specified, Compose implementations MUST reserve all devices that satisfy the requested capabilities. Otherwise, Compose implementations MUST reserve at least the number of devices specified. The value is specified as an integer.
+If `count` is set to `all` or not specified, Compose implementations MUST reserve all devices that satisfy the requested capabilities. Otherwise, Compose implementations MUST reserve at least the number of devices specified. The value is specified as an integer.
 
 ```yml
 deploy:
--- a/deploy.md
+++ b/deploy.md
@@ -161,6 +161,89 @@ services:
 
 `memory` configures a limit or reservation on the amount of memory a container can allocate, set as a string expressing a [byte value](spec.md#specifying-byte-values).
 
+#### devices
+
+`devices` configures reservations of the devices a container can use. It contains a list of reservations, each set as an object with the following parameters: `capabilities`, `driver`, `count`, `device_ids` and `options`.
+
+Devices are reserved using a list of capabilities, making `capabilities` the only required field. A device MUST satisfy all the requested capabilities for a successful reservation.
+
+##### capabilities
+
+`capabilities` are set as a list of strings, expressing both generic and driver specific capabilities.
+The following generic capabilities are recognized today:
+
+- `gpu`: Graphics accelerator
+- `tpu`: AI accelerator
+
+To avoid name clashes, driver specific capabilities MUST be prefixed with the driver name.
+For example, reserving an nVidia CUDA-enabled accelerator might look like this:
+
+```yml
+deploy:
+  resources:
+    reservations:
+      devices:
+        - capabilities: ["nvidia-compute"]
+```
+
+##### driver
+
+A different driver for the reserved device(s) can be requested using `driver` field. The value is specified as a string.
+
+```yml
+deploy:
+  resources:
+    reservations:
+      devices:
+        - capabilities: ["nvidia-compute"]
+          driver: nvidia
+```
+
+##### count
+
+If `count` is set to `-1` or not specified, Compose implementations MUST reserve all devices that satisfy the requested capabilities. Otherwise, Compose implementations MUST reserve at least the number of devices specified. The value is specified as an integer.
+
+```yml
+deploy:
+  resources:
+    reservations:
+      devices:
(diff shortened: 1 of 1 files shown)


EricHripko:
also cc @kohidave

ndeloof (approved):
I like this approach, as it allows both enough abstraction to translate into various Cloud provider but also target local environment or dedicated infrastructure with explicit device IDs.
My +1

deploy.md
```diff
@@ -161,6 +161,89 @@ services:
...
+  resources:
+    reservations:
+      devices:
+        - capabilities: ["nvidia-compute"]
+          driver: nvidia
+```
+
+##### count
+
+If `count` is set to `-1` or not specified, Compose implementations MUST reserve all devices that satisfy the requested capabilities. Otherwise, Compose implementations MUST reserve at least the number of devices specified. The value is specified as an integer.
```
chris-crone, line 204: A bit of a nit but I'm wondering if there's a better way to say "all" than `-1`. The only other place that I can think of in the spec where we do something analogous to "all" is for [deploying global services](https://docs.docker.com/compose/compose-file/#deploy). There we do something like this:

```yaml
services:
  worker:
    image: dockersamples/examplevotingapp_worker
    deploy:
      mode: global
```

Maybe we could do something similar by adding a `mode` field that can have values `partial` or `all` with partial being the default:
```yaml
deploy:
  resources:
    reservations:
      devices:
        - capabilities: ["tpu"]
          mode: all
```

chris-crone (commented):
I like this approach! Only a bit of a nit on how we specify all the devices that match the capabilities

deploy.md
ndeloof: I agree `-1` is a common technical way to define an "unset" value, but is non-intuitive for a user-friendly file format. docker CLI uses "number or `all`" for `--gpu` flag. We could do the same with the `count` attribute.
Another option is indeed to have a separate attribute (which would be exclusive to `count` definition) to select all devices.

EricHripko: I’m leaning towards `count: all`. This is because I’m cautious about the cognitive load of having 3 mutually exclusive ways of requesting a device 🙈

EricHripko: Updated - how does this look @chris-crone?

chris-crone: LGTM!

hangyan (approved):

hangyan (approved):

EricHripko:
Landing since this has 2 approvals (as per Slack conversation) :shipit:

PR #103, opened 2020-09-16
Allow string values for the `cpus` property

aiordache:
Relates to https://github.com/docker/compose/issues/7766 and https://github.com/docker/compose/pull/7768


ndeloof (approved):

EricHripko (approved):
Looks good ✅ In terms of the spec itself:
- [cpus](https://github.com/compose-spec/compose-spec/blob/master/spec.md#cpus) attribute is deprecated, so I guess it's fine to leave it as is (mentioning fractional numbers only)
- [deploy.reservations.cpus](https://github.com/compose-spec/compose-spec/blob/master/deploy.md#cpus) doesn't explicitly mention the type, but the snippet above does show an example using a string

PR #105, opened 2020-09-29
Support nested variable expansion

MarilynFranklin:
Signed-off-by: MarilynFranklin <marilyn.j.franklin@gmail.com>

Add nested parameter expansion to compose specification.

Fixes https://github.com/docker/compose/issues/7794


EricHripko:
Hi @MarilynFranklin - thank you for putting together this PR 👋 Could you please submit an issue to go with it and describe (in a bit more detail) the problem it's solving? Just trying to avoid the "XY problem" situation here since the linked issue states very briefly that you'd like to use nested variable expansion, but doesn't explain what situations such need arises in.

MarilynFranklin:
@EricHripko Sorry for the sparse issue! I've added a new issue here with more detail: https://github.com/compose-spec/compose-spec/issues/113

EricHripko:
Awesome - thank you for this 👍

ndeloof (approved):

PR #107, opened 2020-10-07
Fix schema

robertgzr:

fixes the json-schema definition

https://github.com/cuelang/cue/issues/393
https://github.com/docker/compose/issues/7793

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -435,9 +435,9 @@
                 "additionalProperties": false,
                 "patternProperties": {"^x-": {}}
               }
-            ],
-            "uniqueItems": true
-          }
+            ]
+          },
+          "uniqueItems": true
         },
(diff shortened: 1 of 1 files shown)


EricHripko (approved):
Hi @robertgzr - thank you loads for spotting these issues in the spec and taking the time to improve it for everyone 🙌 As per [JSON Schema docs](https://json-schema.org), these fixes do look right:
- `uniqueItems` should be at the [same level](https://json-schema.org/understanding-json-schema/reference/array.html#uniqueness) as `"type": "array"` 
- `additionalProperties`/`patternProperties` should be at the [same level](https://json-schema.org/understanding-json-schema/reference/object.html#pattern-properties) as `"type": "object"`

Thus, PR looks good ✅

ndeloof (approved):

PR #109, opened 2020-10-19
Add mapping for `DeviceRequest` to compose schema

aiordache:
We need an updated schema with the mapping for `DeviceRequest` to start working on the GPU support in docker-compose.

Relates to https://github.com/compose-spec/compose-spec/pull/100

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -599,7 +599,7 @@
         "type": "object",
         "properties": {
             "capabilities": {"$ref": "#/definitions/list_of_strings"},
-            "count": {"type": "integer"},
+            "count": {"type": ["string", "integer"]},
             "device_ids": {"$ref": "#/definitions/list_of_strings"},
             "driver":{"type": "string"},
             "options":{"$ref": "#/definitions/list_or_dict"}
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -525,7 +525,8 @@
               "properties": {
                 "cpus": {"type": ["number", "string"]},
                 "memory": {"type": "string"},
-                "generic_resources": {"$ref": "#/definitions/generic_resources"}
+                "generic_resources": {"$ref": "#/definitions/generic_resources"},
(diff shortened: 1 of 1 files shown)


EricHripko (commented):
Whoops - sorry, didn't realise that I failed to do the schema changes 🙈 Thanks for picking this up!

I've got a general question first: is the removal of `generic_resources` intentional? If yes, is this a breaking change? I cannot find many (almost any) references to this attribute.

aiordache:
@EricHripko I removed it because I thought it was just a first try to address the GPU mapping. And I didn't find docs about it, maybe I missed something. I think it will break the ECS support as it seemed to rely on it. I would rather update the cloud cli to use the new mapping than having two different mappings for the same resource. 
@ndeloof, @chris-crone what do you think about this?

ndeloof:
removal for `version` comes with a basic rule: 
NEVER REMOVE ANYTHING FROM SCHEMA™

whenever this indeed is a swarm-only thingy, `generic_resources` has been published on compose schema and should stay, even none of the other implementation use it. At most we can mark it DEPRECATED once swarm evolved to adopt this new `devices` attribute

aiordache:
I reverted the removal of `generic_resources`.

ndeloof (approved):

schema/compose-spec.json
```diff
@@ -589,6 +590,23 @@
...
     },
+    
+    "devices": {
+      "id": "#/definitions/devices",
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+            "capabilities": {"$ref": "#/definitions/list_of_strings"},
+            "count": {"type": "integer"},
```
EricHripko, line 601: [count can also be __all__](https://github.com/compose-spec/compose-spec/blob/master/deploy.md#count) to indicate that all devices should be allocated. Could we adjust the schema to allow for that?

EricHripko (commented):
Almost there 🙌 Got one last comment inline 💬

ndeloof (approved):

PR #110, opened 2020-10-21
Introduce `profiles` for services

### LINKED ISSUE #87: Add option to specify services to be started by default
## The Problem / Use Case
`docker-compose` is often used as a simple way to setup and run an application locally for development purposes. For this not always _all_ defined services are required to run - or additional services may be useful for development. But starting the application with a simple `docker-compose up` without explicitly specifying a list of services always starts _all_ defined services.

For example: I like to define an additional service running _phpMyAdmin_ alongside a mysql database service in order to have a convenient way to inspect and manipulate the database in development. But this service should only ever run in development and may even there be optional/only started on-demand. 

Currently the recommended way for this use case is to split the services into multiple files and pass them on the command line
~~~sh
docker-compose -f docker-compose.dev.yml up
~~~

Since the additional services may depend on (e.g. the database service) or extend already defined services this may require to combine multiple files when starting the development environment:
~~~sh
docker-compose -f docker-compose.yml -f docker-compose.dev.yml up
~~~

In an automated environment this is no issue since it just needs to be scripted once. But in a development environment this is a huge inconvenience for the developer:
* the command lines become very long to type instead of a simple `docker-compose up`
* the developer has to remember/lookup which services are defined in which files and pass them as parameters
* the same set of files needs to be passed to _every_ `docker-compose` command since services defined in other files will be ignored, i.e. one can not just stop the application with `docker-compose stop`/`docker-compose down` since the additional services will be ignored

This is a real pain for developers which is why various solutions to this issue were repeatedly requested over years, see linked issues below.

## The desired new feature
There should be a simple way to define the set of "default services" that are started when `docker-compose up` is called without any arguments.
This way additional services can still be defined _in the same_ `docker-compose.yml` file which can then be _started on-demand_:
~~~sh
# start set of "default services"
docker-compose up -d

# start additional service for development
docker-compose up phpmyadmin
~~~

There are various ways the default set could be specified in the `docker-compose.yml` file:
1. provide a file-wide (white)list of services to be started by default, which when not specified defaults to _all_ defined services
2. specify a (black)list of services which should _not_ be started by default but only when explicitly listed or when pulled in as a dependency
3. specify on a per service basis if it should be included in the default set or not; when not specified, the service will be included

## Additional context
This or a similar feature was requested for `docker-compose` repeatedly and in the discussions many more example use cases were explained by the users showing this is a real pain point for them:
docker/compose#6742 (most recent discussion), docker/compose#1896 (main discussion), docker/compose#697, docker/compose#912, docker/compose#1294, docker/compose#1439, docker/compose#1547, docker/compose#2803, docker/compose#3027, docker/compose#3289

Further related issues: docker/compose#1661, docker/compose#1988, docker/compose#2773, docker/compose#3684, docker/compose#4096, docker/compose#4650, docker/compose#6676

### Personal note
I am missing such a feature with `docker-compose` since a very long time. I think it can be implemented with quite reasonable effort but provides an **immense improvement** convenience-wise for local development.
One possible solution I already have implemented which I will explain in the comments.

I'm very committed to making this feature happen whether with my proposed solution or any other different way which I'm also open too. My desired outcome is the improved developer-convenience for the presented use case.
So I am grateful for any feedback and suggestions!

### LINKED ISSUE #97: Proposal: profiles

acran:
Introduce `profiles` in compose schema and specification.
This is based on the PR #98 by @ndeloof and the discussion in #97. It diverges from #98 by:
- limit initial scope to services only
- adds a regex for valid profile names (see https://github.com/compose-spec/compose-spec/pull/98#issuecomment-685732572)
- clarify that the profiles of services explicitly targeted by commands must be automatically enabled (see https://github.com/compose-spec/compose-spec/issues/97#issuecomment-685712571 and https://github.com/compose-spec/compose-spec/issues/97#issuecomment-683398226)
- added another example to demonstrate the last point

Automatically enabling the profiles of explicitly targeted service is the minimal amount of "magic" required here to allow for a sane UX when targeting services with dependencies in non-default profiles. But at the same time it still forces the users to specify a valid compose model where all dependencies share a common profile.

Fixes #97
see also #98
Fixes #87

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -167,6 +167,64 @@ merged are hosted in other folders.
 As some Compose file elements can both be expressed as single strings or complex objects, merges MUST apply to
 the expanded form.
 
+### Profiles
+
+Profiles allow to adjust the Compose application model for various usages and environments. A Compose
+implementation SHOULD allow the user to define a set of active profiles. The exact mechanism is implementation
+specific and MAY include command line flags, environment variables, etc.
+
+The Services top-level element supports a `profiles` attribute to define a list of named profiles. Services without
+a `profiles` attribute set MUST always be enabled. A service MUST be ignored by the Compose
+implementation when none of the listed `profiles` match the active ones, unless the service is
+explicitly targeted by a command. In that case its `profiles` MUST be added to the set of active profiles.
+All other top-level elements are not affected by `profiles` and are always active.
+
+References to other services (by `links`, `extends` or shared resource syntax `service:xxx`) MUST not
+automatically enable a component that would otherwise have been ignored by active profiles. Instead the
+Compose implementation MUST return an error.
+
+#### Illustrative example
+
+```yaml
+services:
+  foo:
+    image: foo
+  bar:
+    image: bar
+    profiles:
+      - test
+  baz:
+    image: baz
+    depends_on:
+      - bar
+    profiles:
+      - test
+  zot:
+    image: zot
+    depends_on:
+      - bar
+    profiles:
+      - debug
+```
+
+- Compose application model parsed with no profile enabled only contains the `foo` service.
+- If profile `test` is enabled, model contains the services `bar` and `baz` which are enabled by the
+  `test` profile and service `foo` which is always enabled.
+- If profile `debug` is enabled, model contains both `foo` and `zot` services, but not `bar` and `baz`
+  and as such the model is invalid regarding the `depends_on` constraint of `zot`.
+- If profiles `debug` and `test` are enabled, model contains all services: `foo`, `bar`, `baz` and `zot`.
+- If Compose implementation is executed with `bar` as explicit service to run, it and the `test` profile
+  will be active even if `test` profile is not enabled _by the user_.
+- If Compose implementation is executed with `baz` as explicit service to run, the service `baz` and the
+  profile `test` will be active and `bar` will be pulled in by the `depends_on` constraint.
+- If Compose implementation is executed with `zot` as explicit service to run, again the model will be
+  invalid regarding the `depends_on` constraint of `zot` since `zot` and `bar` have no common `profiles`
+  listed.
+- If Compose implementation is executed with `zot` as explicit service to run and profile `test` enabled,
+  profile `debug` is automatically enabled and service `bar` is pulled in as a dependency starting both
+  services `zot` and `bar`.
+
+
 ## Version top-level element
 
 Top-level `version` property is defined by the specification for backward compatibility but is only informative. 
@@ -1459,6 +1517,12 @@ ports:
 ### privileged
 
 `privileged` configures the service container to run with elevated privileges. Support and actual impacts are platform-specific.
+
+### profiles
+
+`profiles` defines a list of named profiles for the service to be enabled under. When not set, service is always enabled.
+
+If present, `profiles` SHOULD follow the regex format of `[a-zA-Z0-9][a-zA-Z0-9_.-]+`.
 
 ### pull_policy
 
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -328,6 +328,7 @@
           "uniqueItems": true
         },
         "privileged": {"type": "boolean"},
+        "profiles": {"$ref": "#/definitions/list_of_strings"},
         "pull_policy": {"type": "string", "enum": [
           "always", "never", "if_not_present"
         ]},


spec.md
```diff
@@ -167,6 +167,57 @@ merged are hosted in other folders.
 As some Compose file elements can both be expressed as single strings or complex objects, merges MUST apply to
 the expanded form.
 
+### Profiles
+
+Profiles allow to adjust the Compose application model for various usages and environments. A Compose
+implementation SHOULD allow the user to define a set of active profiles. The exact mechanism is implementation
+specific and MAY include command line flags, environment variables, etc.
+
+The Services top-level element supports a `profiles` attribute to define a list of named profiles. Services without
```
ndeloof, line 176: imho this should not be restricted to `services`, but applied to all top-level elements

ndeloof (approved):

spec.md
acran: For _this_ PR I reduced it to services to keep the scope as small as possible while it also already covers a majority of the use cases with the intention to expand with future PRs :+1:

EricHripko: I think that's a great idea to limit the scope 🙌 To avoid ambiguity, should we clarify what happens to other objects in this case? I guess this can be as simple as specifying that all networks/volumes/etc are treated as always enabled for now.

spec.md
```diff
@@ -167,6 +167,57 @@ merged are hosted in other folders.
...
+a `profiles` attribute set MUST always be enabled. A service MUST be ignored by the compose
+implementation when none of the listed `profiles` match the active ones, unless the service is
+explicitly targeted by a command. In that case its `profiles` MUST be added to the set of active profiles.
+
+References to other services (by `links`, `extends` or shared resource syntax `service:xxx`) MUST not
+automatically enable a component that would otherwise have been ignored by active profiles.
```
EricHripko, line 182: perhaps worth adding that it should also produce an error?

spec.md
```diff
@@ -167,6 +167,57 @@ merged are hosted in other folders.
...
+  zot:
+    image: zot
+    depends_on:
+      - bar
+    profiles:
+      - debug
+```
+
+- Compose application model parsed with no profile enabled only contains the `foo` service.
+- If profile `test` is enabled, model contains the services `foo`, `bar` and `baz`.
```
EricHripko, line 209: I think this is a great example 🎉  Let's explain why each service is pulled in for clarity's sake:
- `foo` = no profiles = always enabled
- `bar` + `baz` = one of profiles matches active profile

spec.md
```diff
@@ -167,6 +167,57 @@ merged are hosted in other folders.
...
+- If profile `debug` is enabled, model contains both `foo` and `zot` services, but not `bar` and `baz`
```
EricHripko, line 210: Same as above

spec.md
```diff
@@ -167,6 +167,57 @@ merged are hosted in other folders.
...
+  and as such the model is invalid regarding the `depends_on` constraint of `zot`.
+- If Compose implementation is executed with `bar` as explicit service to run, it will be active
```
EricHripko, line 212: I think it's worth specifying that `test` profile will be enabled implicitly as well.
It illustrates the spec a bit better and means that, for example, `compose up bar` will also start `baz` (`tests` in the profile) and `foo` (always enabled).

spec.md
```diff
@@ -167,6 +167,57 @@ merged are hosted in other folders.
...
+  even if `test` profile is not enabled by the user.
+- If Compose implementation is executed with `baz` as explicit service to run, the service `baz` will
```
EricHripko, line 214: Same as above

spec.md
```diff
@@ -167,6 +167,57 @@ merged are hosted in other folders.
...
+  be active and `bar` will be pulled in by the `depends_on` constraint.
+- If Compose implementation is executed with `zot` as explicit service to run, again the model will be
+  invalid regarding the `depends_on` constraint of `zot` since `zot` and `bar` have no common `profiles`
+  listed.
+
```
EricHripko, line 219: Given that spec suggests that multiple profiles can be active, I think it's worth adding an example for that

EricHripko (commented):
Looks like an awesome start 🙌 I've left a few comments inline 💬

spec.md
acran: actually no, `compose up bar` will only start `bar` since it has no dependencies to `foo` or `baz` (both would be in the active profiles though, just not started by the command)

acran:
I updated the PR with @EricHripko comments and added/clarified some examples.
But it already feels a bit bloated with that many examples and like over-explaining every single case in detail?

acran:
So, any further comments on the updated PR, or can this be moved forward?

spec.md
EricHripko: This sounds inconsistent. On one hand, the PR says:

> All other top-level elements are not affected by `profiles` and are always active.

On the other hand, you suggest that:

> both would be in the active profiles though, just not started by the command

acran: Sorry, I can not quite follow here:
_All other top-level elements_ is intended to mean _non-service top-level elements_ like _networks_, _volumes_ etc. but `foo` and `baz` are services, so this does not apply here?

With `compose up bar` the services `foo` and `baz` will be _included_ in the compose model for the reasons you mentioned, but they will not be _started_ since only the service `bar` is targeted by the command explicitly^^

EricHripko: Ah sorry, I confused myself here 🙈 Thank you for clarifying!

spec.md
```diff
@@ -167,6 +167,64 @@ merged are hosted in other folders.
...
 the expanded form.
 
+### Profiles
+
+Profiles allow to adjust the Compose application model for various usages and environments. A Compose
+implementation SHOULD allow the user to define a set of active profiles. The exact mechanism is implementation
+specific and MAY include command line flags, environment variables, etc.
+
+The Services top-level element supports a `profiles` attribute to define a list of named profiles. Services without
+a `profiles` attribute set MUST always be enabled. A service MUST be ignored by the compose
```
EricHripko, line 177:
```suggestion
a `profiles` attribute set MUST always be enabled. A service MUST be ignored by the Compose
```
nit: In the spec `Compose` starts with a capital letter

spec.md
EricHripko: `All other top-level elements are not affected by `profiles` and are always active.`

This is now addressed 👍

spec.md
```diff
@@ -167,6 +167,64 @@ merged are hosted in other folders.
...
+  and as such the model is invalid regarding the `depends_on` constraint of `zot`.
+- If profiles `debug` and `test` are enabled, model contains all services: `foo`, `bar`, `baz` and `zot`.
+- If Compose implementation is executed with `bar` as explicit service to run, it and the `test` profile
+  will be active even if `test` profile is not enabled _by the user_.
+- If Compose implementation is executed with `baz` as explicit service to run, the service `baz` and the
+  profile `test` will be active and `bar` will be pulled in by the `depends_on` constraint.
+- If Compose implementation is executed with `zot` as explicit service to run, again the model will be
+  invalid regarding the `depends_on` constraint of `zot` since `zot` and `bar` have no common `profiles`
+  listed.
+- If Compose implementation is executed with `zot` as explicit service to run and peofile `test` enabled,
```
EricHripko, line 223:
```suggestion
- If Compose implementation is executed with `zot` as explicit service to run and profile `test` enabled,
```
Typo

EricHripko (commented):
2 minor wording suggestions - otherwise this looks good to go 🙌

acran:
Thanks @EricHripko 
I now fixed the two typos and also proof-read again, should be fine now :tada:

EricHripko (approved):
Looks great ✅ Thank you for picking up and finishing the initial draft of this proposal 👍

PR #118, opened 2020-12-10
Add missing “build” pull_policy option

gtardif:
Signed-off-by: Guillaume Tardif <guillaume.tardif@gmail.com>

"build" option was missing in https://github.com/compose-spec/compose-spec/pull/59/files#diff-53d1d45e2e85eb9cd869712bc594fa2b30e554973a28c9e363d0dc8f21890bc2R166

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -330,7 +330,7 @@
         "privileged": {"type": "boolean"},
         "profiles": {"$ref": "#/definitions/list_of_strings"},
         "pull_policy": {"type": "string", "enum": [
-          "always", "never", "if_not_present"
+          "always", "never", "if_not_present", "build"
         ]},
         "read_only": {"type": "boolean"},
         "restart": {"type": "string"},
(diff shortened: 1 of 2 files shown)


schema/compose-spec.json
```diff
@@ -330,7 +330,7 @@
         "privileged": {"type": "boolean"},
         "profiles": {"$ref": "#/definitions/list_of_strings"},
         "pull_policy": {"type": "string", "enum": [
-          "always", "never", "if_not_present"
+          "always", "never", "missing", "build"
```
ndeloof, line 333: I don't know any implementation to use it, but as `if_not_present` was in the spec it should not be removed. Maybe we should just declare this as an alias

You also need to define those new values in https://github.com/compose-spec/compose-spec/blob/master/spec.md#pull_policy

gtardif: I'll leave the change on if_not_present for another PR then, my point was to add "build". 
Updated the md specs, thx

ndeloof (approved):

EricHripko (approved):
Looks good ✅ Thank you 👍

PR #125, opened 2020-12-30
Version is optional

### LINKED ISSUE #124: Version is optional
As @elrac [pointed out](https://github.com/compose-spec/compose-spec/issues/13#issuecomment-751972799), the spec still defined `version:` as required even though it optional now (per https://github.com/compose-spec/compose-spec/issues/13).

We should amend the spec to list `version:` as an optional attribute.

**Additional context**
N/A

EricHripko:
This PR removes any unneeded/leftover mentions of version in the spec and ensures that `version:` attribute is defined as optional.

Fixes #124.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -148,7 +148,7 @@ will use a platform-specific lookup mechanism to retrieve runtime values.
 ## Compose file
 
 The Compose file is a [YAML](http://yaml.org/) file defining
-[version](#version) (OPTIONAL),
+[version](#version) (DEPRECATED),
 [services](#service-top-level-element) (REQUIRED),
 [networks](#network-top-level-element),
 [volumes](#volume-top-level-element),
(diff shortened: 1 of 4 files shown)


EricHripko:
also CC @kohidave

spec.md
```diff
@@ -148,7 +148,7 @@ will use a platform-specific lookup mechanism to retrieve runtime values.
 ## Compose file
 
 The Compose file is a [YAML](http://yaml.org/) file defining
-[version](#version) (REQUIRED),
+[version](#version) (OPTIONAL),
```
ndeloof, line 151: We could even mark this as DEPRECATED

ndeloof (approved):

spec.md
EricHripko: Great suggestion! Done 👍

EricHripko:
Thanks for landing this @ndeloof 🙌

PR #127, opened 2020-12-31
Update spec.md with more complete ToC

imthenachoman:
**What this PR does / why we need it**: Easier navigation

**Which issue(s) this PR fixes**: N/A

Fixes #

N/A


ndeloof:
please sign-off your commits or we can't legally accept your contributions
About this PR, please avoid using actual numbers for ordered list, as introducing new flieds would require to change many lines. Markdown will automatically set numbers from `1. xx` on multiple lines. Or just use an unordered list as initial syntax

imthenachoman:
Can you please explain what you mean by "sign-off"? I didn't see any special instructions mentioned in the contributing section but I could have missed it.

EricHripko:
Hi @imthenachoman, see this [SO page](https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for) for more details.

imthenachoman:
Oh. Okay. I was just doing everything through the website and I don't see a sign-off option there. I will pull, sign off, and push when I get time. Although, not sure of value as I suspect most folks don't care about the TOC.

ndeloof:
we could also just adopt https://github.com/thlorenz/doctoc so we don't have to maintain this

PR #131, opened 2021-01-13
Fix broken links

lenzls:
**What this PR does / why we need it**:
This PR will fix several internal links in the "Compose file" section of `spec.md`

**Which issue(s) this PR fixes**:
#130

PR #133, opened 2021-01-21
spec.md: fix links

kenyon:

Fixes links in the document.


kenyon:
I see now that this mostly duplicates #131, but my patch fixes an additional broken link.

ndeloof (approved):

PR #134, opened 2021-01-26
Add a dependsOn condition to wait for successful completion

### LINKED ISSUE #11: investigate support for init container
init container(s) run before service container in sequence and can prepare environment. Think "containerized entrypoint"

We might want to support such init containers in compose spec, with obvious mapping for Kubernetes

syntax proposal and lifecycle document

**Additional context**
See https://github.com/docker/compose/issues/6855

mikesir87:

Updates the spec to add a `dependsOn.condition` that waits for the successful completion of a dependent service. This is important to help support init container workflows.

Fixes #11

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -598,8 +598,8 @@ expressed in the short form.
   - `service_healthy`: specifies that a dependency is expected to be "healthy"
     (as indicated by [healthcheck](#healthcheck)) before starting a dependent
     service.
-  - `service_finished_successfully`: specifies that a dependency is expected to run 
-    to successful completion (as indicated by an exit code of 0)
+  - `service_completed_successfully`: specifies that a dependency is expected to run 
+    to successful completion before starting a dependent service.
 
 Service dependencies cause the following behaviors:
 
--- a/spec.md
+++ b/spec.md
@@ -598,6 +598,8 @@ expressed in the short form.
   - `service_healthy`: specifies that a dependency is expected to be "healthy"
     (as indicated by [healthcheck](#healthcheck)) before starting a dependent
     service.
+  - `service_finished_successfully`: specifies that a dependency is expected to run 
+    to successful completion (as indicated by an exit code of 0)
 
 Service dependencies cause the following behaviors:
 
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -188,7 +188,7 @@
                   "properties": {
                     "condition": {
                       "type": "string",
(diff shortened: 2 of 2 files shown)


ndeloof:
LGTM, two notes:
* I'm not native english but https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-containerdependency.html uses "COMPLETE" and not "FINISHED" for such a state, maybe the former would be a better noun? 
*  if we adopt this I'd expect we also define a `service_completed` value that would not check exit code

mikesir87:
> I'm not native english but https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-containerdependency.html uses "COMPLETE" and not "FINISHED" for such a state, maybe the former would be a better noun?

I had the same debate myself and like `completed` more. So, I switched it to `service_completed_successfully`

> if we adopt this I'd expect we also define a service_completed value that would not check exit code

Honestly, I'm trying to think of the scenario where someone might need that. If the dependent service failed, when would you still want the dependent service to run? Seems quite risky to me, but probably only because I can't think of a use case for it.

ndeloof (approved):
don't forget to update the json schema to reflect this new value

mikesir87:
> don't forget to update the json schema to reflect this new value

And done. Thanks @ndeloof!

spec.md
```diff
@@ -598,6 +598,8 @@ expressed in the short form.
   - `service_healthy`: specifies that a dependency is expected to be "healthy"
     (as indicated by [healthcheck](#healthcheck)) before starting a dependent
     service.
+  - `service_completed_successfully`: specifies that a dependency is expected to run 
+    to successful completion before starting a dependent service.
```
EricHripko, line 602: Apologies if this is overly pedantic, but could we specify that `successful completion` means exiting with code `0`?

EricHripko:
also CC @kohidave

spec.md
ndeloof: we discussed this point during compose-spec meeting and prefered **not** to get into this Linux process implementation detail. Compose model could be applied to distinct things (let's say, VMs, unikernels or WASM payload) which might have a distinct way to notify the platform execution has been successful.

EricHripko: Fair point, resolving 🙂

EricHripko (approved):
Looks good ✅

PR #116, opened 2020-12-04
Updated memory properties with format byte

rtisma:
**What this PR does / why we need it**:
Adds the format `byte` to all `memory` properties. The spec for [memory](https://github.com/compose-spec/compose-spec/blob/68a799e866c2a8ff5d553c3797564282347d2659/deploy.md#memory) indicates the value uses the [byte](https://github.com/compose-spec/compose-spec/blob/68a799e866c2a8ff5d553c3797564282347d2659/spec.md#specifying--byte-values) format

PR #139, opened 2021-02-05
Revert "Updated memory properties with format byte"

ndeloof:
Reverts compose-spec/compose-spec#116


ndeloof:
https://github.com/compose-spec/compose-spec/pull/116 is using `format=byte` but this format is undefined

EricHripko (approved):
Looks good ✅

PR #141, opened 2021-02-18
Fix typo in spec.md

richardkirchofer:
Typo.
"allocate and unassigned host port" -> "allocate any unassigned host port"

Fixes #


EricHripko:
Thank you for putting this together! I believe this should be covered under https://github.com/compose-spec/compose-spec/pull/142, is this the case @richardkirchofer?

EricHripko:
poke @richardkirchofer

richardkirchofer:
You're right! Thanks.

Issue #81, opened 2020-05-22
Support for BuildKit extensions (secrets)


tommie:
### What is the problem you're trying to solve

With multi-stage builds, it's useful to be able to e.g. access an SSH server only from within the build image. Docker now supports that with [BuildKit](https://docs.docker.com/develop/develop-images/build_enhancements/). It adds new command line flags and `Dockerfile` syntax. Compose should allow defining build secrets, just like it supports runtime secrets.

Building using BuildKit is enabled through an environment variable (`DOCKER_BUILDKIT=1`) or a Docker configuration file setting.

Compose already supports runtime secrets, which are files "mounted" inside a container. BuildKit seems very similar, but with some differences:

1. `--ssh`, which simply mounts a Unix socket and sets the `SSH_AUTH_SOCK` environment variable for `RUN` commands that request it.
1. The `dst` flag, which specifies an alternative directory (default is still `/run/secrets/`).
1. Though it's undocumented, it seems [setting owner and mode is supported](https://github.com/moby/buildkit/blob/2769827edcda6d6a2980f94ed3479614745191db/frontend/dockerfile/instructions/commands_runmount.go#L221).

### Describe the solution you'd like

This is a high-level discussion around the requirements of a solution. A more concrete suggestion would probably be more suited for a PR.

Tie in to the existing `secrets`. A [file secret is *defined*](https://github.com/docker/docker-ce/blob/1b27116fcd23fef250f0b3859fb3cbf38ffb3d8e/components/cli/cli/command/image/build_buildkit.go#L442) by

* Type, but only `file` is currently supported
* ID (used to reference the secret in a `RUN`
* Source (the file path)

An [SSH secret is *defined*](https://github.com/docker/docker-ce/blob/1b27116fcd23fef250f0b3859fb3cbf38ffb3d8e/components/cli/cli/command/image/build_buildkit.go#L485) by

* ID
* Agent socket paths

Both of them are referenced/dispatched by specifying

* Source (only for file secrets; implied for SSH)
* Target
* Required (boolean)
* UID
* GID
* Mode

This seems similar enough that it should be possible to extend the `secrets` section, and allow referencing it in an image's `build` section. The SSH type is specific to BuildKit, so parsing code would have to know which type is allowed in build/runtime. Using a separate section for build secrets is a possibility.

[...]

PR #145, opened 2021-02-24
Add support for arbitrary CLI build commands to be passed through to the CLI builder.

zoonage:
**What this PR does / why we need it**: Add support for arbitrary CLI build commands to be passed through to the CLI builder.

**Which issue(s) this PR fixes**: https://github.com/docker/compose/pull/7997

Fixes https://github.com/docker/compose/pull/7997

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -92,7 +92,8 @@
                 "target": {"type": "string"},
                 "shm_size": {"type": ["integer", "string"]},
                 "extra_hosts": {"$ref": "#/definitions/list_or_dict"},
-                "isolation": {"type": "string"}
+                "isolation": {"type": "string"},
+                "extra_cli_build_args": {"$ref": "list_of_strings"}
               },
               "additionalProperties": false,
               "patternProperties": {"^x-": {}}


EricHripko:
Hi @zoonage - thank you for putting this contribution together 👍 It looks like you're aiming to enable SSH secrets, which is being discussed under https://github.com/compose-spec/compose-spec/issues/81. Do you think that your request will be satisfied by the linked issue?

zoonage:
> Hi @zoonage - thank you for putting this contribution together 👍 It looks like you're aiming to enable SSH secrets, which is being discussed under #81. Do you think that your request will be satisfied by the linked issue?

It's turned into more of a "allow arbitrary CLI arguments to be passed through" than something specifically focussed on mounting the SSH agent into the container now

EricHripko:
cc @kohidave too

EricHripko (commented):
Given that we're trying to keep the Compose spec somewhat agnostic of the implementation, I'm concerned about the portability of this addition as it'd essentially lock the Compose file into a single CLI implementation. Would welcome hear thoughts from other maintainers on this though 👀

zoonage:
> Given that we're trying to keep the Compose spec somewhat agnostic of the implementation, I'm concerned about the portability of this addition as it'd essentially lock the Compose file into a single CLI implementation. Would welcome hear thoughts from other maintainers on this though 👀

My thoughts on this are it's hard to strike a balance between being agnostic and having support for new features in Docker as soon as they come out (especially after BuildKit released quite a few things at once).

That said I can change this to specify ssh, secret mounts, and anything else based on what is decided, either way is going to be pretty much the same amount of effort

ndeloof:
My :-1 on this. Passing arbitrary command obviously would locally unlock features but would make it impossible to adapt to another build context. 

Let's imaging we want to provide support for Google Cloud Build. If the build require ssh keys, then it's highly probable they don't just expose buildx command line flags, but offer another way to setup credentials. With adequate abstraction, we can adapt to the infrastructure. 

As a Jenkins contributor I had to fight many times with the lack of abstraction around `withDocker` which basically expose the CLI flags, and make it impossible to adopt an alternate approach.

I'm convince we should *not* hurry here, and just adopt the buildkit features with adequate declarative abstraction. In the meantime, just a reminder we support `x-` extensions if you want to experiment a syntax to introduce ssh support in docker-compose.

ndeloof (changes requested):
:-1 as explained on https://github.com/compose-spec/compose-spec/pull/145#issuecomment-810082609

would better first open an issue on the spec to debate this
also need to update the specification docs if a new field is introduced

zoonage:
> :-1 as explained on [#145 (comment)](https://github.com/compose-spec/compose-spec/pull/145#issuecomment-810082609)
> 
> would better first open an issue on the spec to debate this
> also need to update the specification docs if a new field is introduced

Created https://github.com/compose-spec/compose-spec/issues/157 for discussion, should I leave this PR open for updating after the discussion or close it and open a new one later?

ndeloof:
can be kept open, let's wait for decisions on #157

zoonage:
Closing this out as I don't have capacity to make any progress on this and haven't for a while

PR #153, opened 2021-03-22
Add missing `service.storage_opts`

aiordache:
Missing `storage_opts` service property after we merged the versioned schemas.

https://docs.docker.com/compose/compose-file/compose-file-v2/#storage_opt

Relates to https://github.com/docker/compose/issues/8210


ndeloof (approved):

chris-crone (approved):
LGTM

PR #147, opened 2021-02-26
Proposal: start generating GitHub Pages from the spec

EricHripko:
**What this PR does / why we need it**:
This PR starts an initiative to make the documentation for the Compose spec more user-friendly and easily discoverable.
There are no changes in the content apart from a Jekyll optimisations that enables us to avoid keeping ToC by hand.
I've taken Cayman GH theme and added the Compose logo that's featured on https://compose-spec.io/ and this org.
You can see a preview of this PR under https://erichripko.github.io/compose-spec/.

**Which issue(s) this PR fixes**:
Related to #115.

cc @kohidave

PR #155, opened 2021-03-23
Add a deeper, auto generated ToC for easier, faster navigation

sgpinkus:

Add a deeper auto generated ToC to spec. It's very much needed for such a large spec IMO: faster navigation, and as a good general overview of what's in the spec.

I auto generated this with Atom editor. Can use this is or use preferred auto generator next time you have to update.

No open issue.


EricHripko:
Potentially conflicts with #147

ndeloof:
Would be nice we get a script/automation to generate this ToC

EricHripko:
If we consider GitHub-produced pages, then ToC is already generated from the content (see https://compose-spec.github.io/compose-spec/spec.html).

ndeloof:
Closing as obsolete

PR #163, opened 2021-04-27
introduce create_host_path

ndeloof:
Introduce `create_host_path` as a volume bind options.

as explained on https://github.com/docker/compose-cli/pull/1563#issuecomment-826853661, for legacy reason short syntax binds will let moby container engine create source path on host, while the more recent mount API won't. Introducing this attribute allows to be more explicit on user intents, and can be automatically set when short syntax is used.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1811,6 +1811,9 @@ expressed in the short form.
 - `read_only`: flag to set the volume as read-only
 - `bind`: configure additional bind options
   - `propagation`: the propagation mode used for the bind
+  - `create_host_path`: create a directory at the source path on host if there is nothing present. 
+    Do nothing if there is something present at the path. This is automatically implied by short syntax
+    for backward compatibility with docker-compose legacy.
 - `volume`: configure additional volume options
   - `nocopy`: flag to disable copying of data from a container when a volume is created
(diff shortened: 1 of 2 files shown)


ndeloof:
implementation PR on compose-go: https://github.com/compose-spec/compose-go/pull/132

spec.md
```diff
@@ -1811,6 +1811,7 @@ expressed in the short form.
 - `read_only`: flag to set the volume as read-only
 - `bind`: configure additional bind options
   - `propagation`: the propagation mode used for the bind
+  - `create_host_path`: create source path on host if missing. This is automatically implied by short syntax.
```
gtardif, line 1814: Nit: document this is caused by legacy reasons ?

gtardif (approved):

spec.md
chris-crone:
```suggestion
  - `create_host_path`: create a directory at the source path on host if there is nothing present. Do nothing if there is something present at the path. This is automatically implied by short syntax.
```

chris-crone (commented):
I'd make this more explicit and describe all possible cases.

chris-crone (approved):
LGTM

PR #164, opened 2021-04-27
Add a release process for the specification

### LINKED ISSUE #104: Versioning of the specification for tool makers
While we have removed the user facing version field from the Compose specification, tool makers need some kind of specification versioning so that they can implement their tool against a stable target.

@thaJeztah raised this from the Docker CLI point of view 🤗 

We need to be clear that this is not the same as the old versions of Compose and will not be exposed to the user. It is just something that gives fixed points that tool builders can target. As discussed in the community meeting, this could even be tags based on a date rather than something like semver.

chris-crone:
This PR adds a release process for the specification.

Fixes #104

### DIFF
--- a/RELEASE_PROCESS.md
+++ b/RELEASE_PROCESS.md
@@ -1,0 +1,54 @@
+# Release Process
+
+## Goals
+
+* Do not explicitly expose Compose specification versions to users
+* Keep the process simple and light
+
+## Process
+
+### Requesting a release
+
+To trigger the release process, the requester MUST open a pull request against
+this repository with an updated [CHANGELOG.md](./CHANGELOG.md).
+Anyone can request a release of the specification.
+
+A section MUST be added to the file with the following format:
+
+```markdown
+## date (commit digest)
+
+- list of changes
+```
+
+The date:
+* MUST be of the form `YYYY-MM-DD`
+* SHOULD be the same as when the pull request was opened
+
+The commit digest:
+* MUST be of the same form as that given by the command `git rev-parse HEAD`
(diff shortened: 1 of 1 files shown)


RELEASE_PROCESS.md
```diff
@@ -0,0 +1,54 @@
...
+* MUST be at the point that the release requester would like the release to be
+  made
+
+The list of changes:
+* MUST be a bullet point list
+* MUST list all impactful changes to the specification since the last release
+
+### Approving a release
+
+To approve a release, a simple majority of maintainers MUST approve the pull
```
chris-crone, line 40: Note to reviewers, I put a simple majority here. Do we want to make it less than that?

EricHripko: How do we define maintainers and how many does the majority equal to?

chris-crone: We currently have this list: https://github.com/compose-spec/compose-spec/blob/master/OWNERS.md

We should discuss how we'd like to define maintainership in a community meeting using our [GOVERNANCE doc](https://github.com/compose-spec/compose-spec/blob/master/GOVERNANCE.md) as a starting point.

EricHripko: Sounds good - added it to the agenda of an upcoming meeting 🙂

AkihiroSuda:
What's current status?

I'd also like to see tagged releases of compose-go :heart:

chris-crone:
This is on me: I have an action to finish this PR but I've been struggling to find time 😬 I will try to get this closed out later this week

ndeloof:
Closing as obsolete

PR #165, opened 2021-04-27
Remove Compose on Kubernetes from implementations

chris-crone:
Removes Compose on Kubernetes from the implementations list as it has been [deprecated](https://github.com/docker/compose-on-kubernetes/).


ndeloof (approved):

EricHripko (approved):
Looks great! Happy that we're keeping Compose spec up-to-date 🙂

PR #169, opened 2021-05-04
Add Okteto Stacks to the list of implementations

pchico83:

Add Okteto Stacks to the list of Compose spec implementations

Okteto Stacks are for developers who don't want to deal with the complexities of Kubernetes manifests. Okteto Stacks implement and extend the Compose Specification to make it easy to deploy containers in Kubernetes.

https://okteto.com/docs/reference/stacks

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,9 +30,9 @@ APIs to manage networking services, container deployments and their lifecycles.
 While this offers flexibility to address many operator use cases, it makes
 simple use cases, like the developer use case, more complicated to express than
 they need to be. Projects like [Kompose](https://github.com/kubernetes/kompose)
-show how the simpler Compose model can be translated into Kubernetes API
-payloads and make the Compose file the source of truth for development and
-deployment.
+or [Okteto Stacks](https://okteto.com/docs/reference/stacks) show how the
+simpler Compose model can be translated into Kubernetes API payloads and make
+the Compose file the source of truth for development and deployment.
 
 ### Cloud providers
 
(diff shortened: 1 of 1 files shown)


EricHripko (commented):
Overall, this sounds like a great idea! The only concern that I have is that Otketo extends the spec with an `endpoints` top-level attribute. This means that some tooling will reject it since it's not using `x-` prefix:
```
$ docker-compose config
ERROR: The Compose file './docker-compose.yaml' is invalid because:
'endpoints' does not match any of the regexes: '^x-'
```

pchico83:
> Overall, this sounds like a great idea! The only concern that I have is that Otketo extends the spec with an `endpoints` top-level attribute. This means that some tooling will reject it since it's not using `x-` prefix:
> 
> ```
> $ docker-compose config
> ERROR: The Compose file './docker-compose.yaml' is invalid because:
> 'endpoints' does not match any of the regexes: '^x-'
> ```

@EricHripko should we implement support for `x-endpoints`?

pchico83:
@EricHripko we are working on make all our sample compose compatible, I will update it here when everything is ready

pchico83:
@EricHripko we have updated our docs and samples to be 100% docker-compose compatible.

EricHripko (approved):
Looks great ✅ Thank you for addressing feedback and for demoing Okteto at one of Compose Spec meetings 👍

ndeloof (approved):

PR #173, opened 2021-05-25
document support for port's host_ip in "long syntax"

ndeloof:
Document ability to configure host IP mapping for a port binding when using the "long syntax". Already documented for the "short syntax".

see https://github.com/docker/compose-cli/issues/1701


EricHripko (approved):
Looks great ✅

PR #179, opened 2021-06-19
`missing` is a valid `pull_policy`

### LINKED ISSUE #158: pull_policy schema and spec differ
The spec.md file states that `missing` is a valid value for `pull_policy`, but the schema does not list it in the enum.

Either remove `missing` from `spec.md` or add it to `compose-spec.json` as a valid value for the `pull_policy` enum.

**Additional context**
I haven't checked if adding `missing` to the enum is sufficient for support in docker-compose.

EricHripko:
Looks like `missing` was introduced as the preferred alias for `if_not_present` in https://github.com/compose-spec/compose-spec/pull/136, but the change wasn't reflected in the spec. 

Fixes #158.

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -331,7 +331,7 @@
         "privileged": {"type": "boolean"},
         "profiles": {"$ref": "#/definitions/list_of_strings"},
         "pull_policy": {"type": "string", "enum": [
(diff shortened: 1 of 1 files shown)


EricHripko:
cc @kohidave also

ndeloof (approved):

PR #196, opened 2021-09-03
Clarify specification for external_links

### LINKED ISSUE #195: Spec for external_links is not clear
The documentation for `services.extra_links` reads:

> `external_links` link service containers to services managed outside this Compose application.
> `external_links` define the name of an existing service to retrieve using the platform lookup mechanism.
> An alias of the form `SERVICE:ALIAS` can be specified.

However, the implementation indicates that external links are passed through to the Docker API (after possibly adding an alias if there is no `:` in the input), which expects `CONTAINER_NAME:ALIAS` and does not perform any additional resolution.

Thus, it would be more accurate to say that that this config links service containers to _containers_ managed outside this Compose application. Additionally, the sentence about using the platform lookup mechanism is misleading and should be removed.

**Additional context**
N/A

kendru:
This PR updates the external_links specification to more accurately reflect the behavior of Docker Compose and the Docker engine.

Fixes #195

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -989,9 +989,8 @@ Any other allowed keys in the service definition should be treated as scalars.
 
 ### external_links
 
-`external_links` link service containers to services managed outside this Compose application.
-`external_links` define the name of an existing service to retrieve using the platform lookup mechanism.
-An alias of the form `SERVICE:ALIAS` can be specified.
+`external_links` links service containers to containers managed outside this Compose application.
+An alias of the form `CONTAINER_NAME:ALIAS` can be specified.
 
 ```yml
 external_links:


ndeloof:
The Compose Specification is expected to be generic, so allowing use of `service:xx` with platform to provide some service lookup mechanism is still relevant, while I understand we also need to support docker-compose ability to use an existing container (as the Docker engine without swarm has no concept for "services"). So both should be defined by the spec

kendru:
@ndeloof I understand the desire for this spec to be generic, so maybe `CONTAINER_NAME` is not good terminology. However, since a _service_ is a concept defined by both the Compose spec and by Docker (with Swarm), and this does not correspond to either, I think that a term other than `SERVICE` should be used here. Additionally, the bit about using platform-specific service resolution does not describe the actual behavior of any Compose implementation that I am aware of, so I question whether it belongs in the spec at this point. What are your thoughts?

Issue #204, opened 2021-09-29
tmpfs.size can be define as unit bytes


ndeloof:
**What is the problem you're trying to solve**
Compose file format 2.3 allowed to use `tmpfs.size` with bytes
Compose file format 3.6 only supports integer with constraint `minimum:0`

To apply our goal to support them all, we need to redefine tmpfs.size so both are valid

**Describe the solution you'd like**
a mix of both, but if too complex to be expressed by json schema, better adopt the 2.3 `["integer", "string"]` definition

**Additional context**
https://github.com/docker/compose/issues/8696

PR #205, opened 2021-09-29
define tmpfs.size as string or integer

ndeloof:
Allow tmpfs.size to be defined as unit bytes notation (compatibility with compose file 2.x)

fixes https://github.com/compose-spec/compose-spec/issues/204
see https://github.com/docker/compose/issues/8696

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -427,8 +427,10 @@
                     "type": "object",
                     "properties": {
                       "size": {
-                        "type": "integer",
-                        "minimum": 0
+                        "oneOf": [
+                          {"type": "integer", "minimum": 0},
+                          {"type": "string"}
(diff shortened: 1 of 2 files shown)


schema/compose-spec.json
```diff
@@ -427,8 +427,10 @@
                     "type": "object",
                     "properties": {
                       "size": {
-                        "type": "integer",
-                        "minimum": 0
+                        "oneOf": [
+                          {"type": "integer","minimum": 0},
```
mat007, line 431: nit:
```suggestion
                          {"type": "integer", "minimum": 0},
```

mat007 (approved):
Makes sense!

ulyssessouza (approved):
LGTM

PR #206, opened 2021-10-02
introduce top-level `name`

ndeloof:
introduce top-level `name` to address https://github.com/docker/compose/issues/745

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -227,9 +227,21 @@ SHOULD warn the user. Compose implementations MAY offer options to ignore unknow
 
 ## Name top-level element
 
-Top-level `name` property is defined by the specification as project name to be used if user don't se one explicitly. 
+Top-level `name` property is defined by the specification as project name to be used if user doesn't set one explicitly. 
 Compose implementations MUST offer a way for user to override this name, and SHOULD define a mechanism to compute a
-default project name, to be used only if the top-level `name` element is not set.
+default project name, to be used if the top-level `name` element is not set.
+
+Whenever project name is defined by top-level `name` or by some custom mechanism, it MUST be exposed for 
+[interpolation](#Interpolation) and environment variable resolution as `COMPOSE_PROJECT_NAME`
+
+```yml
+services:
+  foo:
+    image: busybox
+    environment:
+      - COMPOSE_PROJECT_NAME
+    command: echo "I'm running ${COMPOSE_PROJECT_NAME}"
+```
 
 ## Services top-level element
 
--- a/spec.md
+++ b/spec.md
@@ -49,6 +49,8 @@ A **Project** is an individual deployment of an application specification on a p
 resources together and isolate them from other applications or other installation of the same Compose specified application with distinct parameters. A Compose implementation creating resources on a platform MUST prefix resource names by project and
 set the label `com.docker.compose.project`.
 
+Project name can be set explicitely by top-level `name` attribute. Compose implementation MUST offer a way for user to set a custom project name and override this name, so that the same `compose.yaml` file can be deployed twice on the same infrastructure, without changes, by just passing a distinct name.
+
 ### Illustrative example
 
 The following example illustrates Compose specification concepts with a concrete example application. The example is non-normative.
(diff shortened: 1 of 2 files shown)


thaJeztah:
No, I don't think we should do this, and this definitely doesn't address https://github.com/docker/compose/issues/745 (just shifts the problem from directory name or .env to the compose file)

ndeloof:
@thaJeztah then I misunderstood the requirements set on docker/compose#745

ozydingo:
@thaJeztah do you mean because it doesn't persist the project name in a separate file from the docker-compose.yml file, and/or because compose does not save the auto-generated project name into said separate file? This solution does allow persistence of the project name independently of choices that may affect other (not compose related) parts of a project (folder name, checking in a .env file), so I think it's fair to say it covers at least some of the use cases -- I think some clarification on what it's missing would be helpful.

ndeloof:
@thaJeztah ok, reading more carefully your initial request docker/compose#745.

IIUC you're expecting we store "user session state" on filesystem aside the compose.yaml file so that user only has to set `--project-name` once and then all further commands will apply to the same project. We could obviously have docker-compose support some `.docker-compose` file in project directory as an implementation specific alternative to define project name, as well as some additional command flag to create this file. That could be nice (I'm not fully convince this is really useful, but why not ¯\\_(ツ)_/¯ ?

That being said, comments on docker/compose#745 demonstrate we also have user who expect they can hard-code the project name in the compose file to make it fully deterministic. I think we can have both
- default project name defined in compose file
- if not set, default project name determined by implementation (docker-compose uses parent folder)
- project name set by some implementation-specific config file
- project name set explicitly by command

gtardif:
+1 to have an option to define name as part of the compose file. Indeed, it does not need to be the only option, but in many cases defining it inside the compose file is the most intiuitive and natural solution. 
Then, ability to store project name passed by --project-name or other ways when `compose up` can be used to override the name passed in the compose file (that in turn also overrides default naming like based on parent folder). 
Note that compose implementation also uses container labels to store project name, not sure about the interaction with this - but adding the name field in the compose file can be done quite independently of more "runtime" features happening related to storing project name after user executes `compose up`.

gtardif (approved):

spec.md
```diff
@@ -223,6 +225,24 @@ Compose implementations SHOULD validate they can fully parse the Compose file. I
 because the Compose file was written with fields defined by a newer version of the specification, Compose implementations
 SHOULD warn the user. Compose implementations MAY offer options to ignore unknown fields (as defined by ["loose"](#Requirements-and-optional-attributes) mode).
 
+## Name top-level element
+
+Top-level `name` property is defined by the specification as project name to be used if user don't se one explicitly.
```
EricHripko, line 230:
```suggestion
Top-level `name` property is defined by the specification as project name to be used if user don't set one explicitly. 
```

EricHripko (commented):
Tiny typo inline 💬 More general question: how do we expect it to work for cases when multiple Compose files are combined?

ndeloof:
> how do we expect it to work for cases when multiple Compose files are combined

Like other elements : if defined in an override file, will override.

spec.md
```diff
@@ -49,6 +49,8 @@ A **Project** is an individual deployment of an application specification on a p
 resources together and isolate them from other applications or other installation of the same Compose specified application with distinct parameters. A Compose implementation creating resources on a platform MUST prefix resource names by project and
 set the label `com.docker.compose.project`.
 
+Project name can be set explicitely by top-level `name` attribute. Compose implementation MUST offer a way for user to set a custom project name and override this name, so that the same `compose.yaml` file can be deployed twice on the same infrastructure, without changes, by just passing a distinct name.
```
EricHripko, line 52:
```suggestion
Project name can be set explicitly by top-level `name` attribute. Compose implementation MUST offer a way for user to set a custom project name and override this name, so that the same `compose.yaml` file can be deployed twice on the same infrastructure, without changes, by just passing a distinct name.
```

EricHripko (approved):
Looks good (with one typo/suggestion inline) ✅

ulyssessouza (approved):

EricHripko (approved):
Looks great ✅ Very excited about this addition 🎉

PR #207, opened 2021-10-04
Small typo fix

hangyan:
Signed-off-by: Hang Yan <hang.yan@hotmail.com>

Fixes #


EricHripko (commented):
Hi @hangyan 👋 Thank you for looking into this! It doesn't seem like a typo to me. That being said, it might make sense to insert `if` or `whether` into the sentence to make it a bit clearer. For example, `Compose implementations SHOULD validate they whether can fully parse the Compose file`.

EricHripko (approved):
Looks great ✅

ndeloof (approved):

PR #211, opened 2021-11-08
introduce cache-to

ndeloof:
Introduce `cache-to` in the compose specification, to offer a canonical binding with buildkit support for caching.

cc @crazymax @ulyssessouza @tonistiigi

### DIFF
--- a/build.md
+++ b/build.md
@@ -161,7 +161,7 @@ args:
 
 Cache location syntax MUST follow the global format `[NAME|type=TYPE[,KEY=VALUE]]`. Simple `NAME` is actually a shortcut notation for `type=registry,ref=NAME`.
 
-Compose Builder implementations MAY support custom types, the compose specification defineds canonical types which MUST be supported:
+Compose Builder implementations MAY support custom types, the Compose Specification defines canonical types which MUST be supported:
 
 - `registry` to retrieve build cache from an OCI image set by key `ref`
 
@@ -175,7 +175,7 @@ build:
     - type=gha
 ```
 
-Unsupported caches MUST be ignored and not prevent user for building image.
+Unsupported caches MUST be ignored and not prevent user from building image.
 
 ### cache_to
 
@@ -190,6 +190,8 @@ build:
 ```
 
 Cache target is defined using the same `type=TYPE[,KEY=VALUE]` syntax defined by [`cache_from`](#cache_from). 
+
+Unsupported cache target MUST be ignored and not prevent user from building image.
 
 ### extra_hosts
 
--- a/build.md
+++ b/build.md
@@ -157,15 +157,39 @@ args:
 
 ### cache_from
 
-`cache_from` defines a list of images that the Image builder SHOULD use for cache resolution.
+`cache_from` defines a list of sources the Image builder SHOULD use for cache resolution.
+
+Cache location syntax MUST follow the global format `[NAME|type=TYPE[,KEY=VALUE]]`. Simple `NAME` is actually a shortcut notation for `type=registry,ref=NAME`.
+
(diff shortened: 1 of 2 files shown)


build.md
```diff
@@ -157,16 +157,40 @@ args:
 
 ### cache_from
 
-`cache_from` defines a list of images that the Image builder SHOULD uses for cache resolution.
+`cache_from` defines a list of sources the Image builder SHOULD uses for cache resolution.
```
ulyssessouza, line 160:
```suggestion
`cache_from` defines a list of sources the Image builder SHOULD use for cache resolution.
```

EricHripko (commented):
Thank you for putting this together - I got a general question to better understand the use case ❓

I can certainly see the value for a shared cache (e.g., on NFS or in the registry) being referenced by Compose via `cache_from`. What is the use case for specifying `cache_to` in Compose though? Something like a local path especially seems rather machine-specific and possibly better off left to the tooling (e.g., via some CLI interface or `export` command).

ndeloof:
cache_to supports allows to share layers from intermediary images in a multi-stage Dockerfile, which are not pushed with the main image. Also allows to configure cache mechanisms not based on image registry, like github actions cache.

EricHripko:
Thank you for elaborating! I guess there's image, registry and inline caches that aren't machine-specific, strictly speaking 👍

build.md
```diff
@@ -157,16 +157,40 @@ args:
 
 ### cache_from
 
-`cache_from` defines a list of images that the Image builder SHOULD uses for cache resolution.
+`cache_from` defines a list of sources the Image builder SHOULD use for cache resolution.
+
+Cache location syntax MUST follow the global format `[NAME|type=TYPE[,KEY=VALUE]]`. Simple `NAME` is actually a shortcut notation for `type=registry,ref=NAME`.
+
+Compose Builder implementations MAY support custom types, the compose specification defineds canonical types which MUST be supported:
```
EricHripko, line 164:
```suggestion
Compose Builder implementations MAY support custom types, the Compose specification defines canonical types which MUST be supported:
```
typo

build.md
```diff
@@ -157,16 +157,40 @@ args:
...
+
+```yml
+build:
+  context: .
+  cache_to: 
+   - user/app:cache
+   - type=local,dest=path/to/cache
+```
+
+Cache target is defined using the same `type=TYPE[,KEY=VALUE]` syntax defined by [`cache_from`](#cache_from).
```
EricHripko, line 192: Should we (similarly to above) mention that failure to export the cache shouldn't fail the build?

EricHripko (commented):
Looks like a solid addition! Got a quick typo and a question inline 💬

build.md
ndeloof: +1

ndeloof:
@EricHripko fixed, can I get your Approval on this one ?

ulyssessouza (approved):
LGTM

build.md
```diff
@@ -157,16 +157,42 @@ args:
...
+build:
+  context: .
+  cache_to: 
+   - user/app:cache
+   - type=local,dest=path/to/cache
 ```
 
+Cache target is defined using the same `type=TYPE[,KEY=VALUE]` syntax defined by [`cache_from`](#cache_from). 
+
+Unsupported cache target MUST be ignored and not prevent user for building image.
```
EricHripko, line 194:
```suggestion
Unsupported cache target MUST be ignored and not prevent user from building image.
```
typo

EricHripko (approved):
Looks good once the typos inline are sorted ✅ Thank you for seeing this through 👍

PR #213, opened 2021-11-13
Make docker compose v2 the reference implementation

### LINKED ISSUE #212: Deprecate compose-ref
[Compose-ref](https://github.com/compose-spec/compose-ref) is really minimalist and poorly maintained
Compose V2 is released and implemented with the exact same Go runtime and compose-go library for Compose Specification support. 

We should deprecate compose-ref and make Compose V2 the reference implementation for the Compose Specification.

ndeloof:
Make Docker Compose v2 the reference implementation

Close #212

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,12 +55,14 @@ The specification and code is licensed under the Apache 2.0 license found in the
 
 ## Implementations
 
-* [docker-compose](https://github.com/docker/compose)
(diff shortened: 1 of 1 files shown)


chris-crone (approved):
LGTM

ulyssessouza (approved):
LGTM

PR #214, opened 2021-12-02
Add missing deploy.limits.pids (from schema 3.9), deprecate pids_limit

thaJeztah:

compose schema v3.9 added support for `deploy.limits.pids` (through https://github.com/docker/cli/commit/851eeb9639ddd5186c57db0533ede26964ee27cc (https://github.com/docker/cli/pull/2503)), however, work on compose-spec was already in progress, which lead to changes from the 3.9 schema to not be included.

This patch adds the missing option and mars the `services.{name}.pids_limit` as deprecated, similar to comparable limit- and reservation options (`services.{name}.cpus`, `services.{name}.mem_limit`, `services.{name}.mem_reservation`.

relates to https://github.com/moby/moby/issues/43056

### DIFF
--- a/deploy.md
+++ b/deploy.md
@@ -143,6 +143,7 @@ services:
         limits:
           cpus: '0.50'
           memory: 50M
+          pids: 1
         reservations:
           cpus: '0.25'
           memory: 20M
@@ -155,6 +156,10 @@ services:
 #### memory
 
 `memory` configures a limit or reservation on the amount of memory a container can allocate, set as a string expressing a [byte value](spec.md#specifying-byte-values).
+
+#### pids
+
+`pids_limit` tunes a container’s PIDs limit, set as an integer.
 
 #### devices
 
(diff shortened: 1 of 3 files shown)


ndeloof (approved):

EricHripko (approved):
Looks good ✅ Thank you for sorting this inconsistency out!

ulgens:
@thaJeztah Quick irrelevant question:

> compose schema v3.9 added support for

Where is this documented? I mean, I spent hours researching spec details and I couldn't find a single doc (except https://docs.docker.com/compose/compose-file/compose-file-v3/ which defines all versions as "legacy") about differences between different schema versions.

EricHripko:
I believe Docker now includes the [spec](https://docs.docker.com/compose/compose-file/) in their official docs 🤓

ulgens:
@EricHripko As far as I remember it's been around for a while. The problem is it's not clear which spec is supported / implemented and what are the differences between different versions.

EricHripko:
I agree with you 👍 We talked about having MDN-like badges or compatibility tables in the past but never actually gotten around to implementing that in the spec. I think it's something worth revisiting.

PR #222, opened 2022-01-03
published port can be a range

ndeloof:
Makes `published` a string so that it can be set with a range, not just a port number

See https://github.com/docker/compose/issues/8530

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -318,7 +318,7 @@
                   "mode": {"type": "string"},
                   "host_ip": {"type": "string"},
                   "target": {"type": "integer"},
-                  "published": {"type": "integer"},
+                  "published": {"type": ["string", "integer"]},
                   "protocol": {"type": "string"}
                 },
                 "additionalProperties": false,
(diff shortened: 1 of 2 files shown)


rumpl (approved):

schema/compose-spec.json
```diff
@@ -318,7 +318,7 @@
                   "mode": {"type": "string"},
                   "host_ip": {"type": "string"},
                   "target": {"type": "integer"},
-                  "published": {"type": "integer"},
+                  "published": {"type": "string"},
```
mik-laj, line 321:
```suggestion
                  "published": {"type": ["string", "number"]},
```

spec.md
```diff
@@ -1494,7 +1494,7 @@ The long form syntax allows the configuration of additional fields that can't be
 expressed in the short form.
 
 - `target`: the container port
-- `published`: the publicly exposed port
+- `published`: the publicly exposed port. Can be set as a range using xyntax `start-end`, then actual port SHOULD be assigned within this range based on available ports.
```
EricHripko, line 1497:
```suggestion
- `published`: the publicly exposed port. Can be set as a range using syntax `start-end`, then actual port SHOULD be assigned within this range based on available ports.
```
typo

EricHripko (approved):
Got a typo inline, but looks good otherwise ✅ Thank you for picking this up 👍

PR #224, opened 2022-01-11
Fix typo in README.md

bantic:

Small typo fix in the README

N/a


EricHripko:
Thank you for putting this contribution together 👍 There was a bit of a race and this got fixed in https://github.com/compose-spec/compose-spec/pull/225 instead

PR #232, opened 2022-03-02
introduce `no_cache` and `pull` to force a full rebuild

ndeloof:

Introduce `no_cache` and `pull` attributes in the build section.
Those allow to force a full rebuild from sources + sync with registry regarding the base image being used.

I don't expect those to be directly set by end-users in their compose.yml file, but getting them part of the compose model will allow Compose Implementations to offer `--no-cache` and `--pull` to `compose build` (or equivalent) command.


build.md
```diff
@@ -240,6 +240,16 @@ build:
...
 
+### no_cache
+
+`no_cache` disables image builder cache and enforce a full rebuild from source for all image layers. This only
+applies to layers declared in the Dockerfile, base image COULD be retrieved from local image store whenever tag
+has been updated on registry (see #pull).
+
+### pull
+
+`pull` require the image builder to pull a newer image of the base image, even if it is already available in the local image cache.
```
glours, line 251: not only the base image but  all referenced images IHMO, wdyt?

glours (approved):
LGTM

ulyssessouza (approved):

EricHripko (commented):
If we are introducing these primarily for CLI flags and don't expect these to be set by the user, should these be documented at all? Taking it further, should these be part of the schema at all? 🤔

ndeloof:
@EricHripko this is already the case for a bunch of parameter that I don't think user _should_ hardcode in yaml, but some do, because we can't guess all usages... Having this part of the compose spec model makes it more transparent.

EricHripko:
Thank you for elaborating 👍

build.md
```diff
@@ -274,6 +274,17 @@ build:
     - "com.example.label-with-empty-value"
 ```
 
+### no_cache
+
+`no_cache` disables image builder cache and enforce a full rebuild from source for all image layers. This only
+applies to layers declared in the Dockerfile, referenced images COULD be retrieved from local image store whenever tag
+has been updated on registry (see #pull).
```
EricHripko, line 281:
```suggestion
has been updated on registry (see [pull](#pull)).
```
I'm guessing this was meant to be a link.

EricHripko (approved):
Looks great, thank you for looking into this ✅ Left a suggestion inline, feel free to land once it's addressed.

Issue #233, opened 2022-03-02
Improve Service build section


glours:
**What is the problem you're trying to solve**
A lot of issues and pull requests have been opened and asking to improve the build capabilities of the default implementation (Docker Compose).
We already have a bunch of proposals opened and pending in this specification to address some of those requests.

The objective of this issue is to aggregate the existing proposals in one place to facilitate tracking of the implementation progression.

**Describe the solution you'd like**
In the following section we'll define all the fields we want to add to the specification with links to 
- description of the property or link to the existing proposal (if exists)
- Proposal Pull Request link
- Compose-Go PR link
- Compose PR link (optional)

When working on a PR (proposal, compose-go or compose) don't forget to reference this issue

List of the property to add

- cache-to see proposal #120 
  - proposal PR #211 
  - Compose-Go PR https://github.com/compose-spec/compose-go/pull/232
- no-cache see proposal #120
  - proposal PR #232 
  - Compose-Go PR https://github.com/compose-spec/compose-go/pull/233
- platforms see proposal #120 and #197
  - proposal PR #267
  - Compose-Go PR https://github.com/compose-spec/compose-go/pull/282
- pull: `always attempt to pull a newer version of the images used during the build` boolean
  - proposal PR #232 
  - Compose-Go PR https://github.com/compose-spec/compose-go/pull/233
- secrets see proposal #81 
  - proposal PR #238 
  - Compose-Go PR https://github.com/compose-spec/compose-go/pull/236
- ssh: `SSH agent socket or keys to expose to the build, if no value provided should use the SSH_AUTH_SOCK env variable` string , see also #120 
  - proposal PR #234 
  - Compose-Go PR https://github.com/compose-spec/compose-go/pull/234
- tags: `list of the tags associated to the final build image` array, see also 
  - proposal PR #255 
  - Compose-Go PR https://github.com/compose-spec/compose-go/pull/256
- builder-options: `property to allow usage of builder specific build options, for example named context and output property of BuildX` content unspecified by the Compose specification, can be used to enable custom builder features.
  - proposal PR 
  - Compose-Go PR 

**Additional context**

Example using all this modifications
```yaml
services:
  frontend:
    image: awesome/webapp:bar     # image name to be build with a default tag name "bar"
    build:
      context: .
      dockerfile: ./Dockerfile
      args:
        GITHUB_TOKEN: ${GITHUB_TOKEN}
      tags: # define tags that should be associated to the image built
        - ct-addon:foo
[...]

PR #234, opened 2022-03-02
Add ssh to build

glours:
Introduce `ssh` attribute to the `build` section.
This allows the usage of SSH authentification when building an image (i.e., cloning a private repo...)

This PR is part of the #233 proposal

### DIFF
--- a/build.md
+++ b/build.md
@@ -163,7 +163,7 @@ args:
 * `default` - let the builder connect to the ssh-agent.
 * `ID=path` - a key/value definition of an ID and the associated path. Can be either a [PEM](https://en.wikipedia.org/wiki/Privacy-Enhanced_Mail) file, or path to ssh-agent socket
 
-Simple`default` sample
+Simple `default` sample
 ```yaml
 build:
   context: .
@@ -174,7 +174,7 @@ or
 ```yaml
 build:
   context: .
-  ssh: [default]   # mount the default ssh agent
+  ssh: ["default"]   # mount the default ssh agent
 ```
 
 Using a custom id `myproject` with path to a local SSH key:
@@ -185,7 +185,7 @@ build:
     - myproject=~/.ssh/myproject.pem
 ```
 Image builder can then rely on this to mount SSH key during build.
-For illustration, [Buildkit extended syntax](https://github.com/compose-spec/compose-spec/pull/234/%5Bmoby/buildkit@master/frontend/dockerfile/docs/syntax.md#run---mounttypessh%5D(https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md#run---mounttypessh)) can be used to mount ssh key set by ID and access a secured resource:
+For illustration, [BuildKit extended syntax](https://github.com/compose-spec/compose-spec/pull/234/%5Bmoby/buildkit@master/frontend/dockerfile/docs/syntax.md#run---mounttypessh%5D(https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md#run---mounttypessh)) can be used to mount ssh key set by ID and access a secured resource:
 
 `RUN --mount=type=ssh,id=myproject git clone ...`
 
--- a/build.md
+++ b/build.md
@@ -162,14 +162,6 @@ args:
 `ssh` property syntax can be either:
 * `default` - let the builder connect to the ssh-agent.
 * `ID=path` - a key/value definition of an ID and the associated path. Can be either a [PEM](https://en.wikipedia.org/wiki/Privacy-Enhanced_Mail) file, or path to ssh-agent socket
-* `{}` (no value) which is a shortcut of the `default` configuration
-
-`ssh` without any parameter
-```yaml
-build:
-  context: .
-  ssh:   # mount the default ssh agent
-```
 
 Simple`default` sample
 ```yaml
--- a/build.md
+++ b/build.md
@@ -157,7 +157,7 @@ args:
 
 ### ssh
 
-`ssh` defines an SSH authentication that the image builder SHOULD use during image build (e.g., cloning private repository)
+`ssh` defines SSH authentications that the image builder SHOULD use during image build (e.g., cloning private repository)
 
 `ssh` property syntax can be either:
 * `default` - let the builder connect to the ssh-agent.
@@ -167,13 +167,22 @@ Simple`default` sample
 ```yaml
 build:
   context: .
-  ssh: default   # mount the default ssh agent
+  ssh: 
+    - default   # mount the default ssh agent
 ```
+or 
+```yaml
+build:
+  context: .
+  ssh: [default]   # mount the default ssh agent
+```
+
 Using a custom id `myproject` with path to a local SSH key:
 ```yaml
 build:
   context: .
-  ssh: myproject=~/.ssh/myproject.pem
+  ssh: 
+    - myproject=~/.ssh/myproject.pem
 ```
 Image builder can then rely on this to mount SSH key during build.
 For illustration, [Buildkit extended syntax](https://github.com/compose-spec/compose-spec/pull/234/%5Bmoby/buildkit@master/frontend/dockerfile/docs/syntax.md#run---mounttypessh%5D(https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md#run---mounttypessh)) can be used to mount ssh key set by ID and access a secured resource:
--- a/build.md
+++ b/build.md
@@ -161,8 +161,8 @@ args:
 
 `ssh` property syntax can be either:
 * `default` - let the builder connect to the ssh-agent.
-* `ID=socket` - a key/value definition of an ID and the associated socket
-* `ssh` without any value which is a shortcut of the `default` configuration
+* `ID=path` - a key/value definition of an ID and the associated path. Can be either a [PEM](https://en.wikipedia.org/wiki/Privacy-Enhanced_Mail) file, or path to ssh-agent socket
+* `{}` (no value) which is a shortcut of the `default` configuration
 
 `ssh` without any parameter
 ```yaml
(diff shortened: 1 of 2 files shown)


build.md
```diff
@@ -155,6 +155,35 @@ args:
...
 ```
 
+### ssh
+
+`ssh` defines an SSH authentication that the image builder SHOULD use during image build (e.g., cloning private repository)
+
+`ssh` property syntax can be either:
+* `default` - let the builder connect to the ssh-agent.
+* `ID=socket` - a key/value definition of an ID and the associated socket
+* `ssh` without any value which is a shortcut of the `default` configuration
```
ndeloof, line 165:
```suggestion
* `{}` (no value) which is a shortcut of the `default` configuration
```

ndeloof: I suggest we use `{}` for empty as being used in a few other places

build.md
```diff
@@ -155,6 +155,35 @@ args:
   - GIT_COMMIT
 ```
 
+### ssh
+
+`ssh` defines an SSH authentication that the image builder SHOULD use during image build (e.g., cloning private repository)
+
+`ssh` property syntax can be either:
+* `default` - let the builder connect to the ssh-agent.
+* `ID=socket` - a key/value definition of an ID and the associated socket
```
ndeloof, line 164:
```suggestion
* `ID=path` - a key/value definition of an ID and the associated path. Can be either a [PEM](https://en.wikipedia.org/wiki/Privacy-Enhanced_Mail) file, or path to ssh-agent socket
```

build.md
```diff
@@ -155,6 +155,35 @@ args:
...
+  ssh:   # mount the default ssh agent
+```
+
+Simple`default` sample
+```yaml
+build:
+  context: .
+  ssh: default   # mount the default ssh agent
+```
+Using a custom id `myproject` which will be use in the Dockerfile (i.e., `RUN --mount=type=ssh,id=myproject git clone ...`)
```
ndeloof, line 180: example uses buildkit extended syntax: https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md#run---mounttypessh
I'm fine we have implementation-specific examples, but should be noted as such.

I suggest compose.yaml example is followed by an example Dockerfile using buildkit syntax (with link) as an explicit implementation-specific illustration example.

ndeloof (approved):
LGTM, with some minor notes

build.md
levrik: How about using `true` instead? Seems clearer to me as a user but if `{}` is used in other places as well it might be more consistent.

ndeloof (approved):

ulyssessouza (approved):

zaucker:
Do I understand this correctly? This PR just adds the possibility to specify `--ssh` in the `build` section of the docker-compose file, but does not yet pass that to the build command upon running `docker-compose build`?
Just wondering if for my current project I can expect this to become functional soon or if I better don't count on it for the time being.

glours:
@zaucker adding the flag to the `docker compose build` command will be done when we'll implement this specification change in Compose

ndeloof:
implementation in docker/commose should not be a big deal, as we can rely on previous implementation by `buildx bake` as extension fields.

build.md
```diff
@@ -155,6 +155,39 @@ args:
   - GIT_COMMIT
 ```
 
+### ssh
+
+`ssh` defines an SSH authentication that the image builder SHOULD use during image build (e.g., cloning private repository)
+
+`ssh` property syntax can be either:
+* `default` - let the builder connect to the ssh-agent.
```
ndeloof, line 163: `default` might not be a very obvious name, maybe make it more explicit as `ssh-agent`?

levrik: afaik `default` also has to be supplied to the `--ssh` flag of Docker.

build.md
```diff
@@ -155,6 +155,39 @@ args:
...
+* `ID=path` - a key/value definition of an ID and the associated path. Can be either a [PEM](https://en.wikipedia.org/wiki/Privacy-Enhanced_Mail) file, or path to ssh-agent socket
+* `{}` (no value) which is a shortcut of the `default` configuration
```
ndeloof, line 165: sounds a bit weird to me setting `ssh: {}` actually set a value, while - from a yaml point of view - we try to force `null` value, which should reset this field, and entirely omitting this field would have a distinct meaning.

build.md
ndeloof: indeed. We can define an alias then (as we do for `pull: ("always"|"missing"|"never")` for better compatibility with docker CLI

ndeloof: otherwise, I'm also fine we just use `default`, no strong opinion here

build.md
glours: Removed 👍

schema/compose-spec.json
```diff
@@ -91,6 +91,12 @@
                 "context": {"type": "string"},
                 "dockerfile": {"type": "string"},
                 "args": {"$ref": "#/definitions/list_or_dict"},
+                "ssh": {
+                  "oneOf": [
+                    {"type": "string"},
```
ndeloof, line 96: I understand the idea here to allow one-liner `ssh: default` but I wonder we introduce unnecessary complexity with this, and could just define ssh as a `list_or_dict`, which means one can use:
```yaml
ssh: [default]
```
```yaml
ssh: 
  - myproject=~/.ssh/myproject.pem
```
```yaml
ssh: 
  myproject : ~/.ssh/myproject.pem
```
wdyt?

ndeloof: sure it allows something more compact, but with the cost for more confusion. If we allow multiple ssh keys being defined, then let's declare this a list. Otherwise we should allow this for all other places we have lists (volumes, networks, ports, ...)

build.md
```diff
@@ -155,6 +155,31 @@ args:
   - GIT_COMMIT
 ```
 
+### ssh
+
+`ssh` defines an SSH authentication that the image builder SHOULD use during image build (e.g., cloning private repository)
```
ndeloof, line 160: defines ~~an~~ SSH authentication**s**

build.md
```diff
@@ -155,6 +155,33 @@ args:
...
+`ssh` property syntax can be either:
+* `default` - let the builder connect to the ssh-agent.
+* `ID=path` - a key/value definition of an ID and the associated path. Can be either a [PEM](https://en.wikipedia.org/wiki/Privacy-Enhanced_Mail) file, or path to ssh-agent socket
+
+Simple`default` sample
+```yaml
+build:
+  context: .
+  ssh: 
+    default   # mount the default ssh agent
```
ndeloof, line 171:
```suggestion
    - default   # mount the default ssh agent  
```

or 
```
   [default]
```

build.md
```diff
@@ -155,6 +155,40 @@ args:
...
 
+### ssh
+
+`ssh` defines SSH authentications that the image builder SHOULD use during image build (e.g., cloning private repository)
+
+`ssh` property syntax can be either:
+* `default` - let the builder connect to the ssh-agent.
+* `ID=path` - a key/value definition of an ID and the associated path. Can be either a [PEM](https://en.wikipedia.org/wiki/Privacy-Enhanced_Mail) file, or path to ssh-agent socket
+
+Simple`default` sample
```
EricHripko, line 166:
```suggestion
Simple `default` sample
```
typo

build.md
```diff
@@ -155,6 +155,40 @@ args:
...
+build:
+  context: .
+  ssh: 
+    - default   # mount the default ssh agent
+```
+or 
+```yaml
+build:
+  context: .
+  ssh: [default]   # mount the default ssh agent
```
EricHripko, line 177: Should this be `["default"]`? I wouldn't expect that many readers would immediately get that this is an inline list with a single `default` string.

build.md
```diff
@@ -155,6 +155,40 @@ args:
...
+
+Using a custom id `myproject` with path to a local SSH key:
+```yaml
+build:
+  context: .
+  ssh: 
+    - myproject=~/.ssh/myproject.pem
+```
+Image builder can then rely on this to mount SSH key during build.
+For illustration, [Buildkit extended syntax](https://github.com/compose-spec/compose-spec/pull/234/%5Bmoby/buildkit@master/frontend/dockerfile/docs/syntax.md#run---mounttypessh%5D(https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md#run---mounttypessh)) can be used to mount ssh key set by ID and access a secured resource:
```
EricHripko, line 188:
```suggestion
For illustration, [BuildKit extended syntax](https://github.com/compose-spec/compose-spec/pull/234/%5Bmoby/buildkit@master/frontend/dockerfile/docs/syntax.md#run---mounttypessh%5D(https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md#run---mounttypessh)) can be used to mount ssh key set by ID and access a secured resource:
```
nit on casing, feel free to ignore

EricHripko (approved):
Looks awesome ✅ Overjoyed to see SSH support coming to Compose 🙌 I've left a typo fix and some non-blocking comments inline. Feel free to land once the former is addressed.

glours:
Thank you so much @EricHripko for your review, will address you comments before merging

nietzscheson:
This doesn't works for me:

docker compose build --ssh
unknown flag: --ssh

Or:

```bash
### docker-compose.yaml
services:
  core:
    build:
      context: .
      ssh:
        - user_ssh_key=~/.ssh/id_rsa

docker compose build
The Compose file './docker-compose.yaml' is invalid because:
services.core.build contains unsupported option: 'ssh'
```

```bash
docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.7.1)
  compose: Docker Compose (Docker Inc., v2.2.3)
  scan: Docker Scan (Docker Inc., v0.17.0)

Server:
 Containers: 1
  Running: 0
  Paused: 0
  Stopped: 1
 Images: 35
 Server Version: 20.10.12
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 7b11cfaabd73bb80907dd23182b9347b4245eb5d
 runc version: v1.0.2-0-g52b36a2
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.10.76-linuxkit
 Operating System: Docker Desktop
 OSType: linux
 Architecture: x86_64
 CPUs: 6
 Total Memory: 1.939GiB
 Name: docker-desktop
 ID: H3U2:DMWY:AY3Z:NEEF:JQMM:LLUH:MT6Y:OQ2A:3X6V:EHB6:FR6P:UC6Y
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 HTTP Proxy: http.docker.internal:3128
 HTTPS Proxy: http.docker.internal:3128
 No Proxy: hubproxy.docker.internal
 Registry: https://index.docker.io/v1/
 Labels:
 Experimental: false
 Insecure Registries:
  hubproxy.docker.internal:5000
  127.0.0.0/8
 Live Restore Enabled: false
```

PR #221, opened 2022-01-01
Added volume bind option SELinux :z :Z

tymonx:
**What this PR does / why we need it**:

Related with https://github.com/compose-spec/compose-go/pull/213

This feature describe the SELinux field for setting the `:z` or `:Z` bind option for relabeling SELinux label.

References:
- https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label

**Which issue(s) this PR fixes**:

Fixes #191

Signed-off-by: Tymoteusz Blazejczyk <tymoteusz.blazejczyk@tymonx.com>

PR #236, opened 2022-03-04
Revert SELinux

ndeloof:
this revert https://github.com/compose-spec/compose-spec/pull/221

based on discussions on https://github.com/moby/moby/pull/43324, the ability to set SELinux relabeling on mounts is irrelevant, as distinct services could define opposite values with non-deterministic result.
SELinux labels should be set on volume definition, which is not (yet) supported but would be a cleaner way to offer SELinux support. The legacy "bind" API ("short notation") is also wrong from this point of view, but with large adoption we can't change this now.


ndeloof:
cc @tymonx

EricHripko (commented):
The change in the PR seems to be a duplicate of #232 and doesn't match the title. Were the changes on this branch lost? 💬

ndeloof:
my bad, need to re-check my branch

PR #238, opened 2022-03-10
add secrets property to build section

glours:
Add a `secrets` property to the build section which will allow Specification implementations to use secrets during the building step.

Refer to https://github.com/compose-spec/compose-spec/issues/233

### DIFF
--- a/build.md
+++ b/build.md
@@ -301,6 +301,76 @@ build:
   target: prod
 ```
 
+### secrets
+`secrets` grants access to sensitive data defined by [secrets](secrets) on a per-service build basis. Two
+different syntax variants are supported: the short syntax and the long syntax.
+
+Compose implementations MUST report an error if the secret doesn't exist on the platform or isn't defined in the
+[`secrets`](#secrets-top-level-element) section of this Compose file.
+
+#### Short syntax
+
+The short syntax variant only specifies the secret name. This grants the
+container access to the secret and mounts it as read-only to `/run/secrets/<secret_name>`
+within the container. The source name and destination mountpoint are both set
+to the secret name.
+
+The following example uses the short syntax to grant the build of the `frontend` service
+access to the `server-certificate` secret. The value of `server-certificate` is set
+to the contents of the file `./server.cert`.
+
+```yml
+services:
+  frontend:
+    build: 
+      context: .
+      secrets:
+        - server-certificate
+secrets:
+  server-certificate:
+    file: ./server.cert
+```
+
+#### Long syntax
+
+The long syntax provides more granularity in how the secret is created within
+the service's containers.
+
+- `source`: The name of the secret as it exists on the platform.
+- `target`: The name of the file to be mounted in `/run/secrets/` in the
+  service's task containers. Defaults to `source` if not specified.
+- `uid` and `gid`: The numeric UID or GID that owns the file within
+  `/run/secrets/` in the service's task containers. Default value is USER running container.
+- `mode`: The [permissions](http://permissions-calculator.org/) for the file to be mounted in `/run/secrets/`
+  in the service's task containers, in octal notation.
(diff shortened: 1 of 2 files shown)


build.md
```diff
@@ -301,6 +301,76 @@ build:
   target: prod
 ```
 
+### secrets
+`secrets` grants access to sensitive data defined by [secrets](secrets) on a per-service build basis. Two
+different syntax variants are supported: the short syntax and the long syntax.
+
+Compose implementations MUST report an error if the secret doesn't exist on the platform or isn't defined in the
```
ndeloof, line 308:
```suggestion
Compose implementations MUST report an error if the secret isn't defined in the
```
secret MUST be declared in compose file. One can use an existing secret on platform by defining an `external: true` secret

schema/compose-spec.json
```diff
@@ -99,7 +99,27 @@
...
+                        "type": "object",
+                        "properties": {
+                          "source": {"type": "string"},
+                          "target": {"type": "string"},
+                          "uid": {"type": "string"},
+                          "gid": {"type": "string"},
+                          "mode": {"type": "number"}
+                        },
+                        "additionalProperties": false,
+                        "patternProperties": {"^x-": {}}
```
ndeloof, line 118: this model is already declared twice for service configs and secrets, maybe this is a good opportunity to extract it into it's own definition and use a `$ref`

ndeloof (changes requested):
LGTM with some minor adjustments

EricHripko (commented):
Thank you for looking into this! A few high-level questions first ❓ 
- Does this allow for secrets to be optional?
- What would the workflow look like for creating a secret? I'm aware that this crosses somewhat into implementation, but knowing the workflow E2E will help validate the schema changes.

ndeloof:
@EricHripko this is basically a port of `docker build --secret id=mysecret,src=/local/secret` that is consistent with the way docker compose exposes secrets as a bind mount. Obviously, this also can apply to platforms with native secret support, where user might have registered some secret and want them to be available **during the build** as `/run/secret/id`.

EricHripko:
Thank you for confirming 👍 Is the idea that the user calls `compose build --secret ...` then?

ndeloof:
The idea is that use can declare an entry in `secrets` section in yaml file, and use a reference in service.build.secrets so that this secret is exposed during the build process (comparable to the same mechanism used when running services)
secret definition can either include a file (typically, some `~/.secrets-token`) that each user will have to set, or could use an `external: true` secret for platforms with native support for secrets.

schema/compose-spec.json
glours: Yes sure 👍

EricHripko:
That makes sense, thank you for clarifying 👍 One last question, would this still allow for optional secrets?

ndeloof (approved):

ulyssessouza (approved):
LGTM

ndeloof:
> One last question, would this still allow for optional secrets?

The secrets mechanism define in the compose specification has no `optional` capability. This is something we could consider for a future improvements, but then would apply to all secrets, not just the build secrets

EricHripko:
It would be great to allow for optional secrets, as development/CI/production environments can have different secrets (or none at all) even when it comes to build time.

ndeloof:
different secrets makes sense to me (so the ability to provide overrides on compose file, or use paths relative to user's home `~/` to define a secret). I hardly understand how no secret would be used, as this would mean resource can be accessed without a secret? Maybe better discus ability to declare optional secrets in a separate proposal

EricHripko:
Sure, I agree that this can be discussed in a separate thread. Let me get some resources ready to see if there is a substantial justification for such functionality 👍

GhandiJones:
Any update on this?

PR #239, opened 2022-03-15
Fix link to #secrets-top-level-element

seandlg:
Previous link results in a 404. Also, it attempts to point at the wrong location. The actual definition of secrets happens in the [#secrets-top-level-element](https://github.com/compose-spec/compose-spec/blob/master/spec.md#secrets-top-level-element).

None

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1606,7 +1606,7 @@ web:
 
 ### secrets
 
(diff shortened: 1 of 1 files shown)


seandlg:
Will create new PR with signature.

PR #253, opened 2022-04-29
introduce `develop` section

ndeloof:
Introduce a new sub-section `develop` to collect developer workflow and how a Compose Implementation can hook into developer inner-loop to make developer experience better.

update: this proposal has been adjusted to reflect experiment in Docker Compose with `x-develop` and the `watch` command


spec.md
```diff
@@ -262,6 +262,12 @@ the Build section SHOULD be ignored and the Compose file MUST still be considere
 Build support is an OPTIONAL aspect of the Compose specification, and is
 described in detail in the [Build support](build.md) documentation.
 
+Service definition MAY include a Development section, which defines the iner-loop behavior for a developer of the service. The Development section SHOULD be ignored and the Compose file MUST still be considered valid.
```
EricHripko, line 265:
```suggestion
Service definition MAY include a Development section, which defines the inner-loop behavior for a developer of the service. The Development section SHOULD be ignored and the Compose file MUST still be considered valid.
```
typo

development.md
```diff
@@ -0,0 +1,105 @@
+# The Compose Specification - Development support
+{:.no_toc}
+
+*Note:* Development support is an OPTIONAL part of the Compose Specification
```
EricHripko, line 4:
```suggestion
*Note:* Development support is an OPTIONAL part of the Compose Specification.
```
nit: full stop

development.md
```diff
@@ -0,0 +1,105 @@
...
+
+* ToC
+{:toc}
+
+## Introduction
+
+Compose specification is a platform-neutral way to define multi-container applications. A Compose implementation
+focussing on development use-case to run application on local machine will obviously also support some development hooks to improve developer velocity on his local workflow (also known as "inner loop"). The Compose Development specification allows to define how a Compose implementation SHOULD behave to efficiently assist the developer.
```
EricHripko, line 12:
```suggestion
focusing on development use-case to run application on local machine should also support some development hooks to improve developer velocity on their local workflow (also known as "inner loop"). The Compose Development specification allows to define how a Compose implementation SHOULD behave to efficiently assist the developer.
```
some styling suggestions

development.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Definitions
+
+Compose Specification is extended to support an OPTIONAL `development` subsection on services. This section
+defines the development constraints and workflows to be set bu a Compose Implementation to assist the developer
```
EricHripko, line 17:
```suggestion
defines the development constraints and workflows to be set by a Compose Implementation to assist the developer
```
typo

development.md
```diff
@@ -0,0 +1,105 @@
...
+
+  db:
+    image: awesome/database
+    development:
+      watch: sighup
+```
+
+## Developemnt mode definition
+
+The `dev` element define configuration options that are applied by Compose implementations to assist developer during development of a service.
```
EricHripko, line 52:
```suggestion
The `dev` element defines configuration options that are applied by Compose implementations to assist developer during development of a service.
```
typo

development.md
```diff
@@ -0,0 +1,105 @@
...
+
+### watch
+
+`watch` can be specified either as a string containing policy or a detailed structure/
```
EricHripko, line 56:
```suggestion
`watch` can be specified either as a string containing policy or a detailed structure.
```
typo

development.md
```diff
@@ -0,0 +1,105 @@
...
+changes in service source code. "Source code" to watch is identified by the `build.context` section.
+
+While Compose implementations can support custom values, the specification defines:
+- `build`: rebuild service image based on the `build` section and restart the service with updated image
+- `sync`: synchronize source files with container content. `path` attribute is REQUIRED to define the mapping 
+to apply for local files within service `build.context` and container's filesystem.
+- `sighup`: send SIGHUP signal to container, to force a reload configuration/restart. This is typically used 
+for daemon services
+
+If not explicitely configured, and service has a `build` section, then `build` MUST be set as the default policy.
```
EricHripko, line 85:
```suggestion
If not explicitly configured, and service has a `build` section, then `build` MUST be set as the default policy.
```
typo

EricHripko: This is potentially quite a big change in default behaviour. Does this mean that Compose implementations that do support `development` section will assume that all local images need to be rebuilt on each code change?

development.md
```diff
@@ -0,0 +1,105 @@
...
+
+### ignores
+
+The `ignores` attribute can be used to define a list of patterns for paths to be ignored. Any updated file 
+to match a pattern, or belong to a folder to match a pattern, won't trigger service to be re-created. The
```
EricHripko, line 90:
```suggestion
that matches a pattern, or belongs to a folder that matches a pattern, won't trigger service to be re-created. The 
```
something like this IMO provides better readability

development.md
```diff
@@ -0,0 +1,105 @@
...
+
+If not explicitely configured, and service has a `build` section, then `build` MUST be set as the default policy.
+
+### ignores
+
+The `ignores` attribute can be used to define a list of patterns for paths to be ignored. Any updated file 
+to match a pattern, or belong to a folder to match a pattern, won't trigger service to be re-created. The 
+syntax is the same as `.dockerignore` file: 
+
+- `*` match 0 or more characters in a file name.
```
EricHripko, line 93: This list should probably say `matches` instead of `match`

development.md
```diff
@@ -0,0 +1,105 @@
...
+to match a pattern, or belong to a folder to match a pattern, won't trigger service to be re-created. The 
+syntax is the same as `.dockerignore` file: 
+
+- `*` match 0 or more characters in a file name. 
+- `?` macth a single character in file name. 
+- `*/*` match two nested folders with arbitrary names
+- `**` match an arbitrary number of nested folders
+
+If the build context includes a `.dockerignore` file, the patterns in this file are loaded as implicit content
+for the `ignores` file, and values set in the compose model are appended.
```
EricHripko, line 99:
```suggestion
for the `ignores` file, and values set in the Compose model are appended.
```

development.md
```diff
@@ -0,0 +1,105 @@
...
+        policy: build
+```
+
+
+The `watch` attribute defines `update` policy to update a running service as Compose Implementation detects 
+changes in service source code. "Source code" to watch is identified by the `build.context` section.
+
+While Compose implementations can support custom values, the specification defines:
+- `build`: rebuild service image based on the `build` section and restart the service with updated image
+- `sync`: synchronize source files with container content. `path` attribute is REQUIRED to define the mapping
```
EricHripko, line 80: Could you elaborate a bit more on how this interacts with host mounts in `volumes`?

EricHripko (commented):
This looks like a very promising addition - especially the idea potential idea of Compose better supporting remote filesystems out of the box 😍 I've left a few comments and questions inline to better understand this 💬

development.md
ndeloof: this _tells_ compose implementation how it _could_ refresh an image by rebuilding. This is basically the "lowest common denominator". For compiled languages without incremental build capability there's no better choice. Other stacks _might_ offer smoother re-build scenario and we could then embrace those with dedicated policies.

The way a Compose Implementation exposes this to user is TO BE DEFINED by implementation (so that we don't prevent creativity :P). I assume only service image with some detected code change would be rebuilt on code change, and dependent service restarted

development.md
ndeloof: This is intentionally unspecified so implementation can be flexible. Most obvious approach is to add an implicit bind mount for sources. For illustration, a classic nodeJS application uses a COPY directive in Dockerfile to add sources into /app folder. With such a sync directive, `docker compose dev` would add a new bind mount so that local sources override this /app folder and get synced inside container, assuming node runtime manages hot-reload.

EricHripko: Thank you for explaining 👍 My concern is that we are potentially adding another way to mount files into the containers to the spec without necessarily drawing a clear distinction for when users should do volumes and when they should be using `develop` section instead. Is there a use case that this would cover that volumes are not able to?

development.md
```diff
@@ -0,0 +1,111 @@
...
+        trigger: build
+
+  db:
+    image: awesome/database
+    develop:
+      watch: 
+        trigger:
+          strategy: sync
+          # get service to reload configuration
+          signal: SIGHUP
```
EricHripko, line 55: Should this example include a path for what's being synced? According to the definition, build context would be used as a default but one is not defined here.

development.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+## Developemnt mode definition
```
EricHripko, line 58:
```suggestion
## Development mode definition
```
typo

development.md
```diff
@@ -0,0 +1,111 @@
...
+- `*/*` matches two nested folders with arbitrary names
+- `**` matches an arbitrary number of nested folders
+
+If the build context includes a `.dockerignore` file, the patterns in this file are loaded as implicit content
+for the `ignores` file, and values set in the Compose model are appended.
+
+#### quiet_period
+
+If a `quiet_period` is set, Compose Implementations MUST wait at least configured delay before actually refreshing
+service image. Using this allows to avoid service image to be re-created many times in a row as multiple files get
```
EricHripko, line 84: If this attribute applies to `sync` strategy too, then we should make this paragraph a bit more abstract. Currently, it focuses a lot on rebuilding the image.

development.md
```diff
@@ -0,0 +1,111 @@
...
+updated.
+
+#### trigger
+
+`trigger` define the actions to take place as changes have been detected.
```
EricHripko, line 89:
```suggestion
`trigger` defines the actions to take place as changes have been detected.
```
typo

development.md
```diff
@@ -0,0 +1,111 @@
...
+strategies which do not require additional configuration options:
+
+```yaml
+    develop:
+      watch: 
+        # rebuild image and recreate service
+        trigger: build
+```        
+
+Long-syntax allow user to pass additional configuration options, specific to selected strategy. `strategy` attribute
```
EricHripko, line 101:
```suggestion
Long-syntax allows user to pass additional configuration options, specific to selected strategy. `strategy` attribute
```
typo

development.md
```diff
@@ -0,0 +1,111 @@
...
+is REQUIRED:
+
+- `build` strategy will rebuild service image based on the `build` section and restart the service with updated image.
+- `sync` strategy keep existing service contianers running, but can synchronize source files with container content.
```
EricHripko, line 105:
```suggestion
- `sync` strategy keep existing service containers running, but can synchronize source files with container content. 
```
typo

development.md
```diff
@@ -0,0 +1,111 @@
...
+        trigger: build
+```        
+
+Long-syntax allow user to pass additional configuration options, specific to selected strategy. `strategy` attribute
+is REQUIRED:
+
+- `build` strategy will rebuild service image based on the `build` section and restart the service with updated image.
+- `sync` strategy keep existing service contianers running, but can synchronize source files with container content. 
+
+`paths` attribute can be set with `sync` strategy to define local files within service's `build.context` to be synchronized  with
```
EricHripko, line 107: This might be a bit confusing for end users if `source` for this attribute is relative to `build.context` but `source` for volumes is relative to Compose file location.

EricHripko (commented):
Did another pass on this PR and left a few comments and suggestions inline 💬 Thank you for iterating on this!

simonmallet:
Hello! I'm not sure if this is the right place to post comments on the develop file-watch feature ; but that's what it says in the docs at https://docs.docker.com/compose/file-watch/

I have a docker-compose file with 12 services in it.
Most of them will use the watch feature, but not all of them.

For instance I have 2 services that do not mention "x-develop -> watch" in the service, yet when some files get updated, they are rebuilt automatically.

Rebuilding indexer-typescript, azfunc-all-v4-5 after changes were detected:
  - /Users/simonmallet/Sites/cbc/CMS-clone/typescript/current/dist/libs/rcgraph/src/lib/interfaces/withShortStoryTileTemplate.d.ts
  - /Users/simonmallet/Sites/cbc/CMS-clone/typescript/current/dist/libs/rcgraph/src/lib/valueObjects/rightType.js

`azfunc-all-v4-5:
    build:
      context: ../typescript/current
      dockerfile: apps/content-notification/docker-local/Dockerfile.azfunc-all
    env_file:
      - .env
      - ../typescript/current/.env.local
    environment:
      environment: ${Environment}
      DeployedEnvironmentWithIndex: ${EnvironmentWithIndex}
    depends_on:
      - "mongo"
      - "elasticsearch"`

I'd like to be able to use docker-compose alpha watch command and that it does NOT watch apps that do not mention the x-develop option. Or an alternative would be to specify x-develop -> watch and use a pre-defined option : none.

When I build an app it writes a lot of files to the dist folder and the docker-compose alpha watch command goes in frenzy rebuilding everything forever using all my hard-disk space.

ndeloof:
@simonmallet github.com/docker/compse would be the right place for this as `watch` at this point is experimental

tr00st:
(Assuming this is the right place for this discussion, as per the announcement blog post linking back here...)

Related to @EricHripko's comments re: watch.source paths being relative to the build.context, if this is the intention, it would be useful to specify a reference to build.additional_contexts to allow syncing of code outside the main build context. Something like a `from_context` entry (named based on `COPY --from=` syntax from the Dockerfile, though "watch from" probably makes less sense semantically).

ndeloof:
@tr00st see https://github.com/compose-spec/compose-spec/pull/253#issuecomment-1548231579 ;)

tr00st:
> @tr00st see https://github.com/compose-spec/compose-spec/pull/253#issuecomment-1548231579 ;)

Aah. Apologies, thought that was specifically a bug, hadn't read all the way to the end. I'll try and find the right place to report the dodgy advice for feedback in docs and blog posts ;)

UserQuant:
Very nice feature, works well if the changes occur in one file. But, if you copy several files or a folder with files, the terminal freezes and nothing is copied. This also happens if you change the development branch. Action rebuild does not rebuild the image and restart the container, it just syncs with the container. Config i used for compose just like in [documentation](https://docs.docker.com/compose/file-watch/)

nemchik:
> _There is another action which just as important, and FAR simpler than those: "restart only"._

Just to clarify, you're looking for a watch with a pull and recreate, not just a restart, correct?

I know this thread isn't about 3rd party tooling, but just to mention a few ways I've been doing this to make life easier:

- [watchtower](https://github.com/containrrr/watchtower) pretty much does exactly this, although I would encourage caution with configuring it to globally update all containers (better to be specific and configure it using labels)
- [renovate](https://docs.renovatebot.com/modules/manager/docker-compose/) can keep a compose file in a git repo up to date, and can do so with pull requests so incoming changes are visible rather than blindly applied. Updating a git repository with a compose file is only a part of the puzzle though, and you'll need to consider how to deploy that version controlled compose to your system (there are many options for this, potentially even the new watch options coming to compose).

Sevi7:
Hi, in the [documentation](https://docs.docker.com/compose/file-watch/) it says:

> Rules from .dockerignore apply
>    - Use include / exclude to override

However, **include** attribute does not work. Thanks

ndeloof:
@Sevi7 this PR is an early proposal for `develop` section, it doesn't match the current syntax used by `watch` in Docker Compose. Please open an issue on https://github.com/docker/compose

ndeloof:
rebased and updated to reflect Docker Compose `x-develop` experimental feature

glours (approved):
Sounds good to me as the first iteration of this new section 🥳

develop.md
```diff
@@ -0,0 +1,91 @@
...
+services:
+  frontend:
+    image: awesome/webapp
+    build: ./webapp
+    develop:
+      watch: 
+        # sync static content
+        - path: ./src
+          action: sync
+          target: /var/www
```
milas, line 35: Build context is `./webapp`, so the path should probably be under it.

```suggestion
        - path: ./webapp/src
          action: sync
          target: /app
```

develop.md
```diff
@@ -0,0 +1,91 @@
...
+          ignore:
+            - node_modules/
+
+  backend:
+    image: awesome/backend
+    build: ./backend
+    develop:
+      watch: 
+        # rebuild image and recreate service
+        - path: ./src
```
milas, line 45:
```suggestion
        - path: ./backend
```

develop.md
```diff
@@ -0,0 +1,91 @@
...
+- `?` matches a single character in file name. 
+- `*/*` matches two nested folders with arbitrary names
+- `**` matches an arbitrary number of nested folders
+
+If the build context includes a `.dockerignore` file, the patterns in this file are loaded as implicit content
+for the `ignores` file, and values set in the Compose model are appended.
+
+##### path
+
+`path` attribute defines the path to source code (relative to `build.context`) to monitor for changes. Updates to any file
```
milas, line 84:
```suggestion
`path` attribute defines the path to source code (relative to the project directory) to monitor for changes. Updates to any file
```

develop.md
```diff
@@ -0,0 +1,91 @@
...
+inside path, which doesn't match any `ignore` rule, will trigger configured action.
+
+##### target
+
+`target` attribute onbly applies when `action` is configured for `sync`. Files within `path` with changes will be synchronized
```
milas, line 89:
```suggestion
`target` attribute only applies when `action` is configured for `sync`. Files within `path` with changes will be synchronized
```

schema/compose-spec.json
```diff
@@ -463,6 +464,27 @@
...
+      "type": ["object", "null"],
+      "properties": {
+        "watch": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "ignore": {"type": "array", "items": {"type": "string"}},
+              "path": {"type": "string"},
+              "action": {"type": "string", "enum": ["rebuilt", "sync"]},
```
milas, line 478:
```suggestion
              "action": {"type": "string", "enum": ["rebuild", "sync"]},
```

milas (changes requested):
Few typo fixes & adjustment re: the path relativity

develop.md
ndeloof: Really ? When did we change this ?

ulyssessouza (approved):
LGTM 🥳

glours:
@lonix1 please open a new proposal (issue) for your need. We had to ship a first version of this new section of the spec to move forward and now we're in position to start incremental evolution of it.

Avihais12344:
Hello, from what I have read [here](https://docs.docker.com/compose/file-watch/#feedback), this is the place to report bugs and give feedback on the new feautre: `docker compose watch` command.

I am loving this feature, and it works pretty fine, but I have a problem. When I exit the watch command by using ctrl + c, or by using the command: `docker compose down --timeout=0 --volumes --remove-orphans`, there's still a prosess that holds some lock file. As I get the error:

```bash
cannt take exclusive lock for project "my_project": process with PID 11324 is still running
```

What can I do? The only solution I have found is to reset the computer.
The output of the: `docker-compose version` command:

```bash
Docker Compose version v2.22.0-desktop.2
```

Thanks in ahead for any help :)

ndeloof:
@Avihais12344 this is an implementation issue, please report on https://github.com/docker/compose

mhabsaoui:
Hi,

First thanks for this very nice feature, it is a lot anf **granular** user friendly than bind mounting ;-)

I used it and it's simply awsome :+1: 
 
I just wanted to make a small feedback about the logs :  with `docker compose watch` command we get the compose watch mode logs only, not those related to service container app.

As the watch command can't be launched in detached mode like `composer up`, you have to open another terminal to launch the command `docker compose -f compose.dev.yml logs -f -t` to get them !

AFAIK, there isn't any detached ou logs flag dedicated to **watch** command, as for **up** one.

Thanks.

mhabsaoui:
> Hi,
> 
> First thanks for this very nice feature, it is a lot **granular** and user friendly than bind mounting ;-)
> 
> I used it and it's simply awsome 👍
> 
> I just wanted to make a small feedback about the logs : with `docker compose watch` command we get the compose watch mode logs only, not those related to service container app.
> 
> As the watch command can't be launched in detached mode like `docker compose up` and doesn't give the hand back, you have to open another terminal to launch the command `docker compose -f compose.dev.yml logs -f -t` to get them ! PS: See screenshot below...
> 
> AFAIK, there isn't any detached ou logs flag dedicated to **watch** command, as for **up** one. It would be very usefull though :-)
> 
> Thanks.
> 
> image

Just my 2 cents, could be the compose up command with a watch flag : `docker compose up --watch`

Thanks

MarcusElevait:
Do i always have to have a build section for the watch mode?

So for example we have a simple js application and i want to run this one in a container watching file changes for hot reload. 
The easiest thing i could imagine would be to just use a simple node-alpine image and copy all the files into it and run an npm command to start the app.

If i understand the watch feature correctly i now need a Dockerfile that covers all the stuff i mentioned above and use this one in the compose file.

But couldn't it make sense, that i just define the node image in the compose file and the copying is done by the watch section.

So instead of this in the compose file:
```yaml
    web:
        build:
            context: .
            dockerfile: ./Dockerfile
        command: npm start
        develop:
            watch:
                - action: sync
                  path: .
                  target: /src
``` 

where the Dockerfile looks like this:
```
FROM node:18.16.0-alpine

WORKDIR /app

COPY . .
```

it could be just this in the compose file:
```yaml
    web:
        image: node
        command: npm start
        develop:
            watch:
                - action: sync
                  path: .
                  target: /src
```

Because now i always have to build the image.

ndeloof:
@MarcusElevait the solution you describe would mean application can't run correctly without `watch` to manage file sync, i.e. `docker compose up` would just fail.
`watch` is designed to make the inner-loop faster, without the need to stop-rebuild-restart on any code change, but still the application should be well defined without this feature enabled.

For your specific scenario, you could use inlined dockerfile:
```yaml
    web:
        build:
            context: .
            dockerfile-inline: |
FROM node
COPY . /src
        command: npm start
        develop:
            watch:
                - action: sync
                  path: .
                  target: /src
```

MarcusElevait:
What i've done now is this:
```yaml
  web:
        image: node
        command: npm start
        volumes:
            - ./:/src
```

the downsides i see currently when using watch instead are:

- have to build the image with the copy command, which takes longer
- cannot use a service name in the `docker compose watch` command to just start and watch this service and it's dependend services and ignore other services in the compose file (even tough this should be possible according to documentation)
- it is way slower and more resource intensive with a bigger application
- when shutting down the watch container, making some file changes and then starting the container again, it does not rebuild the image and therefore also does not recognize the changes that were made during the downtime.

g0t4:
+1 @MarcusElevait `dotnet watch` needs a `--build` arg at a minimum (if not as the default behavior) given its intended use as dev tooling.

Chokhonelidze:
docker compose watch is always rebuilding image while docker compose up --build is skipping rebuild.
is there any way to make it work as docker compose up --build ?

maciej-wakula-opuscapita:
related: [#418](https://github.com/compose-spec/compose-spec/issues/418)

aruku:
I submitted this feedback a few days ago to a Google Docs Form I found, but I just discovered this seems to be the official way, so I'll leave here too:
Why not adding a bidirectional sync setting? When I run a container for development, all my binaries are there, including the package manager, so that is where I add and remove packages. But with the current workflow, I have to do docker cp to bring the updated dependencies and lock files down to the host. This option would eliminate that step.

i-am-unknown-81514525:
Would it be possible to just `restart` instead of `sync+restart` as sometime those file might already been updated through volume mount and so it do not have to be sync again

ndeloof:
watch configuration is designed to avoid you set a bind mount on your own, so doesn't really make sense

i-am-unknown-81514525:
does it?
(I was just finding this feature interesting because sync take less time than rebuilding the whole image, and restart would be faster than sync+restart as well

leandrodesouzadev:
Hello there!
When I have a "container_name" directive on my compose file, and using the "rebuild" option on the watch action, i get an error looking like:
```log
time="2024-07-23T09:23:26-03:00" level=warning msg="Error handling changed files for service celeryworker: Error response from daemon: Conflict. The container name \"/0eb11b45dc0b_local_django\" is already in use by container \"c8590d3b762fafcd826f7964516442d08ad3c3edb56c67db8e926d8f31d112de\". You have to remove (or rename) that container to be able to reuse that name."
```

Compose file (partial):
```yaml

services:
  django: &django
    build:
      context: .
      dockerfile: ./compose/local/django/Dockerfile
    image: local_django
    container_name: local_django
    depends_on:
      - postgres
      - redis
    env_file:
      - path: ./.envs/.local/.django
      - path: ./.envs/.local/.postgres
      - path: .env
        required: false
    ports:
      - '8001:8000'
    command: /start
    extra_hosts:
      - "host.docker.internal:host-gateway"
    develop:
      watch:
        - action: sync
          path: .
          target: /app
          ignore: 
            - ./requirements
        - action: rebuild
          path: requirements
```

The goal here was to rebuild the application when the "requirements" folder content was changed.

derekrprice:
> does it? (I was just finding this feature interesting because sync take less time than rebuilding the whole image, and restart would be faster than sync+restart as well for instance, I just want `restart` when config file change

This is my use case for a plain "restart" action.  I want to leave the bind in place.  Then I don't need to worry about changing users to get write permissions to files that I normally leave read-only in my production environment.  When my `pnpm run build:watch` rebuilds my Typescript sources into executable JS, I just want the docker container to restart.  It's hosting a websocket server, not a hot-reloading development webserver, so it needs a full restart too.  The simple sync from the example in the docs won't do the trick.

development.md
Ricardonacif: This is still unclear for me.

ei-grad:
Just found this functionality, and also concerned about the missing plain "restart" action, which will restart container if path is changed.

Also, it will be convenient to watch files by mask and allow to specify several paths per rule (like, allow `path` to be a list for the `rebuild` action).

ei-grad:
It also would be nice if there was a documented way to disable "w Enable watch" message in `docker compose up <service>`.

ndeloof:
@ei-grad see https://github.com/docker/compose/issues/11446
About compose navigation menu, this is specific to docker compose, not compose-spec. You can disable by running `docker compose up --menu=false` or set `COMPOSE_MENU=false` in your environment

PR #254, opened 2022-05-02
no sign-off needed for organisation members

glours:
it will let us to easily merge PRs when we accept modifications directly from the GitHub UI

### DIFF
--- a/.github/dco.yaml
+++ b/.github/dco.yaml
@@ -1,0 +1,2 @@
+require:
+  members: false


ndeloof (approved):

glours:
cc @justincormack any objections? ATM only admin can merge PR when we commit suggestion from the GitHub UI because those commits aren't sign-off

justincormack:
No, we need signoffs.

PR #272, opened 2022-08-12
Mark published port as string

### LINKED ISSUE #262: Spec inconsistency between ports short and long syntax on port range
There is a discrepancy in the [ports](https://github.com/compose-spec/compose-spec/blob/master/spec.md#ports) section of the specification: the short syntax does not seem to support host port range only (e.g. something like "8080-8090:80") and it is even explicitly stated that "Host and container **MUST** use equivalent ranges."

But, on another hand, the long syntax does allow for such expression and it clearly appears in one of the given examples:
```
ports:
  - target: 80
    host_ip: 127.0.0.1
    published: 8000-9000
    protocol: tcp
    mode: host
```
Which is equivalent to short syntax "8000-9000:80". Such change has been introduced with PR https://github.com/compose-spec/compose-spec/pull/222

As Docker CLI currently allows the use of host port range only via `-p <host port range>:<container port>` (and so Docker Compose), and to get a consistent short/long syntax in the spec, update short syntax documentation as follows:

* remove statement "Host and container MUST use equivalent ranges."
* update short syntax examples with a host port range entry ("9090-9099:80" for e.g.)

**Additional context**
Such clarification will allow to ask for host port range only support on Podman Compose (and so Podman) as it currently fails at starting a Compose manifest using such configuration (both Podman Compose and Podman fail with error _"Error: host and container port ranges have different lengths"_).

glours:
Update the `ports.published` attribut description to explicitly indicate that a string format is expected, as the specification use the Yaml unquoted string feature some user may implicitly assume that only integer value are allowed.

Fixes #262 and link to https://github.com/docker/compose/issues/9306

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1509,6 +1509,7 @@ ports:
   - "8000:8000"
   - "9090-9091:8080-8081"
   - "49100:22"
+  - "8000-9000:80"
   - "127.0.0.1:8001:8001"
   - "127.0.0.1:5000-5010:5000-5010"
   - "6060:6060/udp"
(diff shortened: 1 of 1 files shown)


ulyssessouza (approved):

spec.md
```diff
@@ -1520,7 +1521,7 @@ The long form syntax allows the configuration of additional fields that can't be
 expressed in the short form.
 
 - `target`: the container port
-- `published`: the publicly exposed port. Can be set as a range using syntax `start-end`, then actual port SHOULD be assigned within this range based on available ports.
+- `published`: the publicly exposed port. Can be set as a range using syntax `start-end`,so it is defined as a string, then actual port SHOULD be assigned within this range based on available ports.
```
EricHripko, line 1524:
```suggestion
- `published`: the publicly exposed port. Can be set as a range using syntax `start-end`, so it is defined as a string, then actual port SHOULD be assigned within this range based on available ports.
```

EricHripko (approved):
Looks good ✅ Thank you for improving this - left a tiny suggestion inline 💬

ndeloof (approved):

Issue #276, opened 2022-10-05
The Compose Specification - secrets - Long syntax - Missing information


spisarau:
`target: The name of the file to be mounted in /run/secrets/ in the service's task containers. Defaults to source if not specified.`
From this string  is not clear what happen if I provide name of the file like this **/server.cert**. Will it store in memory or not.

glours:
@spisarau 
There is an [issue](https://github.com/compose-spec/compose-spec/issues/161) to track this problem and a [PR](https://github.com/compose-spec/compose-spec/pull/281) is ready to clarify it.
Does it answer your question?

glours:
Closed by #281

Issue #95, opened 2020-08-15
Proposal: custom secrets mounted target


hangyan:
**What is the problem you're trying to solve**
allow users to choose which dir they can mount their secets

**Describe the solution you'd like**
By default, it shoud be `/run/secrets/` as the old ways, but a new `target` field can be set to specified as the mount target

**Additional context**
related to #93

Enquier:
This is definitely confusing since at minimum the Docker tooling has supported custom secret paths since 2017x. It also increases the complexity of container implementations to have to account for /run/secrets rather than the default location they look for the secret.

glours:
Fixed by #281

PR #281, opened 2022-10-25
Clarify secret.target absolute path syntax

### LINKED ISSUE #161: secret.target: explicitly allow or disallow absolute path

[The current spec](https://github.com/compose-spec/compose-spec/blob/4423eba8c0926bf0dc57b116b2426d57e024ffb1/spec.md#long-syntax-3) does not seem to allow an absolute path like `/foo/bar` as a secret target:

> target: The name of the file to be mounted in /run/secrets/ in the service's task containers. Defaults to source if not specified.

However, this seems to have been allowed in `docker-compose`.

Revise the spec to explicitly allow (or disallow) absolute path.

epuronta:

The current spec does not define what happens if one specifies an absolute target path when specifying secrets. Specifying an absolute path is supported by Docker (at least) and works just fine.

Fixes #161 (also #95 and #276)

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1645,7 +1645,7 @@ the service's containers.
 
 - `source`: The name of the secret as it exists on the platform.
 - `target`: The name of the file to be mounted in `/run/secrets/` in the
-  service's task containers. Defaults to `source` if not specified.
+  service's task container, or absolute path of the file if an alternate location is required. Defaults to `source` if not specified.
 - `uid` and `gid`: The numeric UID or GID that owns the file within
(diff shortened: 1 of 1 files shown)


glours:
@epuronta can you sign-off you commit? If you don't know how to do that, just follow [those instructions](https://github.com/compose-spec/compose-spec/pull/281/checks?check_run_id=9088616350)

epuronta:
@glours updated the source branch with a signed-off commit. DCO check seems to have passed now.

Thank you for taking care of this!

ndeloof (approved):

PR #282, opened 2022-11-09
Proposal: Pass UID and GID to compose

dmitryuk:
Please check description here https://github.com/compose-spec/compose-go/pull/299


EricHripko (commented):
Thank you for proposing this 👍 I expect introducing this in a portable way would be quite a challenge, some questions that come to mind are:
- What should this do on other *nix-like systems (like MacOS)?
- Is `-1` a reasonable default or will it just result in container (`ENTRYPOINT` or mount) immediately exploding on startup because it's not a valid UID/GID?
- How does this fit it with Windows containers that Compose also supports?

milas (commented):
Having felt the pain of bind mounts many times myself, I understand the utility in this!

However, I'm not actually sure if I think this is a good fit for the specification.

First, the closest similar part of the spec is `COMPOSE_PROJECT_NAME`, which is available for use in variable interpolation:
> Whenever project name is defined by top-level name or by some custom mechanism, it MUST be exposed for [interpolation](https://github.com/compose-spec/compose-spec/blob/master/12-interpolation.md) and environment variable resolution as `COMPOSE_PROJECT_NAME`

That is fairly similar, though it's very "meta", i.e. Compose-specific rather than system-related, which I think is a useful distinction.

Given that it's system-related, it seems much more implementation specific, e.g. we've already noted that it won't work on Windows.

It also raises questions around spec handling in other situations where you wouldn't necessarily have a logical UID/GID (e.g. a linter, CI system).

From that perspective, I think it's a mix of portability and "wrong-level" concern to be part of the spec for me.

However, I think it's reasonable to support in `docker/compose`:
* docker/compose#7853
> 💁🏻 Disclaimer: I'm a maintainer of Docker Compose, and I would happily push to prioritize the feature there.

Environment variables are already "non-portable" in a sense, and are an extensibility mechanism, so I don't see any harm in a particular implementation providing extra variables about the system / Git commit / CI job / etc.

dmitryuk:
@milas, thanks for your answer so much.
You mean that we don't need to have changes in compose-spec repo, because already have agreements about environment variables, that's right?
One thing that UID, GID aren't environments, but shell variables.
Also please take a look into exiting PR to compose-go https://github.com/compose-spec/compose-go/pull/299 I've made before. 
I can close this PR as not needed, but https://github.com/compose-spec/compose-go/pull/299 should be continued

ndeloof:
comparable PR https://github.com/compose-spec/compose-go/pull/572 was rejected for portability concerns

PR #285, opened 2022-12-06
extra_host can be declared as a list or mapping

ndeloof:
Allows use of a list or a mapping to declare extra_hosts, as supported by Docker Compose v1 and v2

Fixes https://github.com/docker/compose/issues/10038


laurazard (approved):
LGTM

PR #286, opened 2022-12-06
introduce support for OCI artifacts for volumes

ndeloof:
introduce `artifact` attribute for volume definition, for Compose implementation to initialize newly created volume based on OCI artifact content

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -657,6 +657,7 @@
       "type": ["object", "null"],
       "properties": {
         "name": {"type": "string"},
+        "content": "$ref": "#/definitions/content"},
         "driver": {"type": "string"},
         "driver_opts": {
           "type": "object",
@@ -679,6 +680,18 @@
       },
       "additionalProperties": false,
       "patternProperties": {"^x-": {}}
+    },
+
+    "content": {
+      "id": "#/definitions/content",
+      "type": ["object", "null"],
+      "properties": {
+        "artifact": {"type": "string"},
+        "file": {"type": "string"},
+        "pull_policy": {"type": "string", "enum": [
+          "always", "missing", "never"
+        ]}
+      }
     },
 
     "secret": {
--- a/spec.md
+++ b/spec.md
@@ -2127,6 +2127,46 @@ volumes:
 An entry under the top-level `volumes` key can be empty, in which case it uses the platform's default configuration for
 creating a volume. Optionally, you can configure it with the following keys:
 
+### content
+
+Specify an initial content to set in the volume once created and before being mounted
+by serice containers.
+
+```yml
+volumes:
+  example:
+    content:
+      artifact: "registry-1.docker.io/docker/demo:0.0.1"
+      pull_policy: always
+```
+
+Child attributes `artifact` and `file` are exclusive. A Compose implementation SHOULD report an error if both are set.
+
+#### artifact
+
+Specify reference to an [OCI artifact](https://github.com/opencontainers/artifacts) used to populate volume when created.
+When set, a Compose implementation MUST pull the OCI artifact from reference and populate newly created volume with content.
+
+`artifact` MUST define an OCI registry reference for an OCI artifact with media-type `application/vnd.docker.volume.v1+tar.gz`.
+If media-type doesn't match, a Compose implementation SHOULD report an error to the user.
+
+#### pull_policy
+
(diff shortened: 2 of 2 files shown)


spec.md
```diff
@@ -2127,6 +2127,20 @@ volumes:
 An entry under the top-level `volumes` key can be empty, in which case it uses the platform's default configuration for
 creating a volume. Optionally, you can configure it with the following keys:
 
+### artifact
+
+Specify reference to an [OCI artifact](https://github.com/opencontainers/artifacts) used to populate volume when created.
+When set, a Compose implementation MUST pull the OCI artifact from reference and populate newly created volume with content.
```
milas, line 2133: Should we use the same behavior as `image`? Default pull policy of "missing" with option to specify via long syntax?

https://github.com/compose-spec/compose-spec/blob/1eb16a20675324b35258b74b8375b48399893eab/spec.md?plain=1#L1556-L1568

```yaml
volumes:
  example:
    oci:
      artifact: "registry-1.docker.io/docker/demo:0.0.1"
      pull_policy: "always"
```

spec.md
```diff
@@ -2127,6 +2127,20 @@ volumes:
...
+
+```yml
+volumes:
+  example:
+    artifact: "registry-1.docker.io/docker/demo:0.0.1"
```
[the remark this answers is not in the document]
ndeloof: I don't think so. We don't refer to images as "OCI" while they also follow the OCI distribution spec.
BUT, thinking twice about this, maybe we should introduce a more generic support for `content` as initial volume content, which _could_ be set by an OCI artifact, but maybe also other sources (a local tar.gz?) :

```
volumes:
  example:
    content: 
       artifact: "registry-1.docker.io/docker/demo:0.0.1"
       pull_policy: "always"

  another_example:
    content: 
       file: ./test_data.tar.gz
```

spec.md
ndeloof: A challenge as we introduce `pull_policy` here is that Docker engine has no local cache for OCI artifacts, so Compose will need to manage his own, or just ignore this field. Maybe better keep it for a future evolution of the spec as alternatives get developed ?

milas:
> maybe we should introduce a more generic support for content as initial volume content, which could be set by an OCI artifact, but maybe also other sources (a local tar.gz?)

I think this is really interesting! Perhaps instead of this being a new volume _type_, we could have something in the `volumes` section?

```yaml
services:
  foo:
    image: nginx
    volumes:
      - webdata:/usr/share/nginx/html

volumes:
  webdata:
    content:
      artifact: registry-1.docker.io/docker/demo:0.0.1
      # OR
      path: ./foo
      # OR
      path: ./bar.tar
```

(cc @nicks @nicksieger "data loading" has definitely come up as a request in Tilt a bunch of times)

ndeloof:
> Perhaps instead of this being a new volume type, we could have something in the volumes section?

This is exactly my proposal, see https://github.com/ndeloof/compose-spec/blob/volume_artifact/spec.md#content

milas:
D'oh! I knew that and then confused myself today thinking it was somehow part of the `service` spec 🤦

schema/compose-spec.json
```diff
@@ -681,6 +682,18 @@
       "patternProperties": {"^x-": {}}
     },
 
+    "content": {
+      "id": "#/definitions/content",
+      "type": ["object", "null"],
```
EricHripko, line 687: How come we're specifying `null` here?

EricHripko (commented):
Thank you for looking into this 🙌 I can envision the use case behind sharing volumes via OCI artefacts, but struggling a bit with local archives. Could you elaborate on how you envision this to be used?

schema/compose-spec.json
laurazard: @ndeloof could you TAL?

spec.md
```diff
@@ -2127,6 +2127,46 @@ volumes:
 An entry under the top-level `volumes` key can be empty, in which case it uses the platform's default configuration for
 creating a volume. Optionally, you can configure it with the following keys:
 
+### content
+
+Specify an initial content to set in the volume once created and before being mounted
+by serice containers.
```
laurazard, line 2133:
```suggestion
by service containers.
```

laurazard (approved):
LGTM, just left some nits 🥳

ndeloof:
This proposal is postponed as such support should better be implemented by container runtime, otherwise users will be able to create a volume from OCI by compose, but not have an equivalent `docker volume create` command.

PR #289, opened 2022-12-10
introduce support for service hooks

ndeloof:
Introduce support for hooks command to run after container started, and/or before container is (intentionaly) stopped

Typical use-case for this is to 
- import initial dataset in a service
- prepare service for shutdown with a dedicated command.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1113,6 +1113,25 @@ healthcheck:
   disable: true
 ```
 
+### hooks
+
+`hooks` declares additional commands to run inside service container(s) during lifecycle. 
+Hooks are commands, and as such can be defined either as a single string, or as a list, in a manner similar to service's [#command](command).
+
+```yml
+services:
+  database:
+    image: foosql
+    hooks:
+      on_create: ./generate_server_certificate.sh
+      on_stop: ["/bin/sh", "echo", "Good bye!"]
+```
+
+- `on_start` hooks defines command a Compose implementation MUST execute after container has been started. There's no guarantee this hook will execute before container's ENTRYPOINT completed. This hook is executed any time a service container is started or restarted.
+- `on_stop` hooks defines command a Compose implementation MUST execute before container get stopped.
+
+If hook execution fails (return code != 0) a Compose implementation SHOULD consider the container to be in a failing state, and stop the Compose application before reporting an error to the user.
+
 ### hostname
 
 `hostname` declares a custom host name to use for the service container. MUST be a valid RFC 1123 hostname.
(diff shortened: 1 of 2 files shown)


ndeloof:
> restart another container when a container is recreated

if you use `depends_on` ou already have such use-case supported (while, recreating containers, not just restarting)

> use “docker compose exec” to run a command within the same container that just started [ie: a minor permission change on a file after start or the use cases of the original poster]

this is exactly what `on_start` is about.


I'm reticent to giving option to run commands out of the container: while I understand this allows many advanced usages, this brings portability and security issues.

EricHripko:
I agree re: @ndeloof's point on running commands on the host - mostly on the portability front. Compose spec today is somewhat detached from the implementation and the generic "execute command on host" feature would make the spec less portable (by tying it to a specific Compose CLI and relying on specific CLI utilities existing on the host).

spec.md
```diff
@@ -1113,6 +1113,25 @@ healthcheck:
...
+
+`hooks` declares additional commands to run inside service container(s) during lifecycle. 
+Hooks are commands, and as such can be defined either as a single string, or as a list, in a manner similar to service's [#command](command).
+
+```yml
+services:
+  database:
+    image: foosql
+    hooks:
+      on_create: ./generate_server_certificate.sh
```
EricHripko, line 1126: Is this meant to say `on_start` instead?

EricHripko (commented):
Exciting - thank you for looking into this! Left a quick comment inline 💬 I'd also suggest naming the hooks something like `after_start` and `before_stop` to better indicate (through naming) when the hooks are executed, as hinted by @michaelkrieger.

laurazard (approved):
Looks great to me ✅ ! Super happy about this change, and I think it'll simplify a ton of very common workflows. Echo @EricHripko and @ndeloof's sentiments re: execution on the host.

JS-E:
Any update on this? Awaiting a small PR tweak

ulyssessouza (changes requested):
I think there are some steps missing:
- A global `pre` and `post`. That is analogous to a unit tests that have "BeforeEverything" and "AfterEverithing" steps.
- `pre_create` and `post_create`. Actually every `pre` and `post` for each state in the state machine of the container.
- Not running in the service container, but in another one. Or at least it shouldn't be the default.  (DEFINITELY NOT RUNNING IN THE HOST!)

ndeloof:
@ulyssessouza this proposal is based on kuke lifecycle hooks
it is only focussing on running _inside_ service container, but you can read https://github.com/compose-spec/compose-spec/issues/291 for init-containers running before `start`
About debate using entrypoint script vs post-start hook, better ask k8s users how and why they use this feature vs writing a more complex entrypoint. Same question could apply to pre-stop vs application code to react to SIGTERM to shutdown nicely

PR #292, opened 2022-12-16
introduce support for cgroup=host|private

### LINKED ISSUE #148: Missing composer specification equivalent for --cgroupns=[host|private]/API1.41
On hosts with "only" a unified cgroup v2 controller hierarchy (that is, a "pure" cgroups v2 hierarchy) Docker now defaults (unless configured otherwise) to automatically creating private cgroup namespaces for created containers. The rationale here is to reduce leakage of potentially sensitive information about the cgroup hierarchy and configuration of the host, and thus other containers. Still, there are valid usecases to allow only specific containers to use the initial (host) cgroup namespace, especially for system diagnosis containers. For instance, the well-known [cAdvisor](https://github.com/google/cadvisor) ("*provides container users an understanding of the resource usage and performance characteristics of their running containers*").

For such usecases, docker run introduced the CLI flag --cgroupns=[host|private] as of Docker-CE 20.0.0 and API 1.41.

Unfortunately, the current compose specification doesn't provide any means to controlling cgroup namespace creation or sharing. This makes it impossible to use system diagnosis containers, such as cAdvisor, ... on cgroup v2 unified hierarchy-only systems anymore with compose-based deployment tools.

This situation would be solved by adding a cgroup field to service objects that would allow to specify the values of either "private" or "host", mirroring today's CLI flag functionality for handling the cgroup namespace creation for new containers.

```yaml
services:
  foo:
    pid: host
    cgroup: host # might be "cgroupns" instead if parity with CLI flag is sought after, but breaks symmetry with pid namespaces.
```

Nota bene: mirror [docker/compose issue #8167](https://github.com/docker/compose/issues/8167)

ndeloof:
Fixes #148
see https://github.com/docker/compose/issues/8167

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -140,6 +140,7 @@
         },
         "cap_add": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
         "cap_drop": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
+        "cgroup": {"type": "string", "enum": ["host", "private"]},
         "cgroup_parent": {"type": "string"},
         "command": {
           "oneOf": [
--- a/spec.md
+++ b/spec.md
@@ -405,6 +405,14 @@ cap_drop:
   - SYS_ADMIN
 ```
 
+### cgroup
+
+`cgroup` specifies the cgroup namespace to join. When unset, it is container runtime decision to
+select cgroup namespace to use, if supported.
+
+- `host`: Run the container in the Container runtime cgroup namespace
+- `private`: Run the container in its own private cgroup namespace
+
 ### cgroup_parent
 
 `cgroup_parent` specifies an OPTIONAL parent [cgroup](http://man7.org/linux/man-pages/man7/cgroups.7.html) for the container.


ulyssessouza (approved):

thediveo:
https://github.com/compose-spec/compose-spec/issues/148#issuecomment-834639159 lists the docker CLI allowed values which include an empty string:

```
docker run --help | grep cgroup
      --cgroupns string                Cgroup namespace to use (host|private)
                                       'host':    Run the container in the Docker host's cgroup namespace
                                       'private': Run the container in its own private cgroup namespace
                                       '':        Use the cgroup namespace as configured by the
                                                  default-cgroupns-mode option on the daemon (default)
```

Would this be also needed to be carried over to the compose world?

ndeloof:
@thediveo `cgroup: ""` would probably better be defined in a compose file by not setting `cgroup` attribute. Then the actual behavior would be left to the container runtime (out of the scope of the compose spec) which is indeed  `default-cgroupns-mode option` on docked, but could be a different choice with alternative runtimes.

BorderCloud:
+1

glours (approved):

thediveo:
This has also finally landed in the documentation: https://docs.docker.com/compose/compose-file/05-services/#cgroup

PR #294, opened 2022-12-19
add privileged attribute to build section

### LINKED ISSUE #279: Support privileged builds in Compose
The spec currently doesn't offer means to declare privileged builds. This functionality has, however, somewhat recently been enabled by the release of BuildKit build engine. As such, this is related to https://github.com/compose-spec/compose-spec/issues/157.

It'd be great if we could enable privileged builds from the [build](https://github.com/compose-spec/compose-spec/blob/master/build.md) configuration. This could look similar to below:
```yaml
build:
  context: .
  privileged: true
```
Such configuration would enable a privileged build with BuildKit and allow usage of instructions like [RUN --security=insecure
](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/reference.md#run---securityinsecure).

**Additional context**
Some build systems make use of sandboxing (e.g., filesystem isolation or process namespacing) in order to achieve reproducible/isolated builds. Making use of these inside a container requires privileged mode, which is difficult to enable at image build time in a `Dockerfile`. Examples include:
- [nix image](https://hub.docker.com/r/nixos/nix), which currently has to disable sandboxing for `RUN nix-build` instruction to work in the `Dockerfile`;
- [buildstream images](https://gitlab.com/BuildStream/buildstream-docker-images/-/blob/master/USING.md), which ship wrapper scripts for the host to ensure that privileged mode is enabled when containers are spawned.

Whilst one could disable sandboxing or mimic what such build systems do with `Dockerfile` instructions, this requires notable effort and forks the development/testing workflow away from "production build" workflow. Supporting privileged builds in Compose should enable developers to instead reproduce the real build of the software and effectively get a production-grade container image from Compose CLI invocation. This leads to many benefits, as developers then get to enjoy great UX of Compose tooling in order to develop/test their software locally or in CI.

glours:
Signed-off-by: Guillaume Lours <705411+glours@users.noreply.github.com>

Add support of privileged mode when building service images

Fixes #279

### DIFF
--- a/build.md
+++ b/build.md
@@ -250,6 +250,16 @@ configuration, which means for Linux `/etc/hosts` will get extra lines:
 `isolation` specifies a build’s container isolation technology. Like [isolation](spec.md#isolation) supported values
 are platform-specific.
 
+### privileged
+
+`privileged` configures the service image to build with elevated privileges. Support and actual impacts are platform-specific.
+
+```yml
+build:
+  context: .
+  privileged: true
+```
+
 ### labels
 
 `labels` add metadata to the resulting image. `labels` can be set either as an array or a map.
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -102,6 +102,7 @@
                 "shm_size": {"type": ["integer", "string"]},
                 "extra_hosts": {"$ref": "#/definitions/list_or_dict"},
                 "isolation": {"type": "string"},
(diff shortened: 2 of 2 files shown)


ndeloof (approved):

laurazard (approved):

EricHripko:
This is super exciting 😍 Thank you for adding this to the spec!

Issue #277, opened 2022-10-06
No UTS namespace option


evsasha:
**What is the problem you're trying to solve**
No [UTS docker setting](https://docs.docker.com/engine/reference/run/#uts-settings---uts) in docker-compose.

**Describe the solution you'd like**
Аdd this option to the compose format.

**Additional context**
Related to docker/compose#8408
Related to docker/compose#3338

EricHripko:
Hi @evsasha 👋 Thank you for sharing this feedback - this seems like a great addition for the spec!

man4red:
I would like to see that option too

PR #297, opened 2022-12-23
Add support of `uts` in service definition

laurazard:
Signed-off-by: Laura Brehm <laurabrehm@hey.com>

Adds support of `uts` mode setting to service definition

Fixes https://github.com/compose-spec/compose-spec/issues/277
related to https://github.com/docker/compose/issues/3338
and https://github.com/docker/compose/issues/8408

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1177,6 +1177,17 @@ which MUST be implemented as described if supported:
     ipc: "service:[service name]"
 ```
 
+### uts
+
+`uts` configures the UTS namespace mode set for the service container. When unspecified
+it is the runtime's decision to assign a UTS namespace, if supported. Available values are:
+
+- `'host'` which results in the container using the same UTS namespace as the host.
+
+```yml
+    uts: "host"
+```
+
 ### isolation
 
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -1177,6 +1177,18 @@ which MUST be implemented as described if supported:
     ipc: "service:[service name]"
 ```
 
+### uts
+
+`uts` configures the UTS namespace mode set for the service container. Available values are:
+
+- `''` which results in the container having it's own UTS namespace.
```
ndeloof, line 1184: to align with other attributes, empty value should not be documented (i.e: this is up to container runtime to assign an UTS namespace, depending on it's implementation details and configuration).
see https://github.com/compose-spec/compose-spec/blob/master/spec.md#cgroup for illustration

ndeloof (commented):
schema.json also need to be updated to reflect this change

ndeloof:
would it make sense we also support the `service:xx` syntax, as we do for other namespaces?

ndeloof (approved):

PR #300, opened 2023-01-20
clarify support for Git URLs a build context

ndeloof:
Clarify support for Git URLs as build context and supported syntax

see https://github.com/compose-spec/compose-go/pull/346


EricHripko (approved):
Looks great ✅ Thank you for adding this - this should make reusing container build files in "microservice per repository" scenario considerably easier 🤩

laurazard (approved):

PR #305, opened 2023-02-14
introduce `restart` for depends_on

ndeloof:
introduce additional `restart` attribute on `depends_on`.
This let user explicitly opt-in for service to be restarted after a dependency has been restarted.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -609,6 +609,10 @@ starting a dependent service.
 The long form syntax enables the configuration of additional fields that can't be
 expressed in the short form.
 
+- `restart`: when `true` a Compose implementation MUST restart this service after it updated the dependency service.
+  This applies to an explicit restart controlled by a Compose operation, and excludes automated restart by the container runtime
+  after container died.
+
 - `condition`: condition under which dependency is considered satisfied
   - `service_started`: is an equivalent of the short syntax described above
   - `service_healthy`: specifies that a dependency is expected to be "healthy"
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -609,6 +609,8 @@ starting a dependent service.
 The long form syntax enables the configuration of additional fields that can't be
 expressed in the short form.
 
+- `restart`: when `true` a Compose implementation MUST restart service after dependent service has been restarted.
```
milas, line 612:
```suggestion
- `restart`: when `true` a Compose implementation MUST restart this service after updates to the dependency service.
```

(English is horrible here, there's no non-confusing way to talk about dependencies. 😡) Looking at this section, though, we refer to the "parent" service as "dependency service" and the "child" service as dependent, so trying to be consistent

Also, I'm not crazy about my change from "has been restarted" to "after updates" - we probably need another sentence of clarification here to be explicit that the restart will only happen as part of a Compose operation, e.g. if `parent` dies and gets auto-restarted by Docker Engine, nobody is going to restart `child` in that case. That is, this is a "deployment" (`up`) level thing, there's no background daemon monitoring

milas (approved):
LGTM, I nitpicked the wording but no concerns on the actual spec/schema

spec.md
ndeloof: We could make it more explicit as 
" a Compose implementation MUST restart this service after it updated the dependency service. " so that users don't expect some magic to happen :)

laurazard (approved):
LGTM

Issue #306, opened 2023-02-15
Add support for configuring additional build contexts


laurazard:
**What is the problem you're trying to solve**

BuildKit added the `--build-context` argument to enable users to provide additional build contexts during a build. More documentation about this can be found [here](https://docs.docker.com/engine/reference/commandline/buildx_build/#build-context).

**Describe the solution you'd like**

Being able to configure additional build contexts in a Compose file, such as:

```yaml
services:
  a:
    build:
      context: .
      build_contexts:
        - project=[a directory/image/etc.]
```

Used together with a Dockerfile such as:

```Dockerfile
# syntax=docker/dockerfile:1
FROM alpine:latest
COPY --from=project . /
```

**Additional context**
Add any other context about the feature request here.

- https://github.com/docker/compose/issues/9961 => https://github.com/microsoft/vscode-remote-release/issues/7305

ndeloof:
We should just use `contexts` as this is in the `build` section already

laurazard:
Sounds good! I'll edit the proposal.

ndeloof:
`additional_contexts` or `named_contexts` could indeed avoid confusion.
Also would probably better use of a mapping vs the `name=path` syntax introduced by buildx as a command line flag doesn't have much flexibility (we could support both, as we do in a few places already)

PR #307, opened 2023-02-16
Add `additional_contexts` to `build` section

laurazard:

Adds support for configuring additional/named contexts in the `build` section.

Fixes https://github.com/compose-spec/compose-spec/issues/306

### DIFF
--- a/build.md
+++ b/build.md
@@ -227,6 +227,36 @@ Cache target is defined using the same `type=TYPE[,KEY=VALUE]` syntax defined by
 
 Unsupported cache target MUST be ignored and not prevent user from building image.
 
+### additional_contexts
+
+`additional_contexts` defines a list of named contexts the image builder SHOULD use during image build.
+
+`additional_contexts` can be a mapping or a list:
+
+```yml
+build:
+  context: .
+  additional_contexts:
+    - resources=/path/to/resources
+    - app=docker-image://my-app:latest
+    - source=https://github.com/myuser/project.git
+```
+
+```yml
+build:
+  context: .
+  additional_contexts:
+    resources: /path/to/resources
+    app: docker-image://my-app:latest
+    source: https://github.com/myuser/project.git
+```
+
+When used as a list, the syntax should follow the `NAME=VALUE` format, where `VALUE` is a string. Validation beyond that is the responsibility of the image builder (and is builder specific).
+
+The Compose implementation SHOULD warn the user if the image builder does not support additional contexts and MAY list the unused contexts.
+
+Illustrative examples of how this is used in Buildx can be found [here](https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#-additional-build-contexts---build-context).
+
 ### extra_hosts
 
 `extra_hosts` adds hostname mappings at build-time. Use the same syntax as [extra_hosts](spec.md#extra_hosts).
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -97,6 +97,7 @@
                 "cache_from": {"type": "array", "items": {"type": "string"}},
                 "cache_to": {"type": "array", "items": {"type": "string"}},
                 "no_cache": {"type": "boolean"},
+                "additional_contexts": {"$ref": "#/definitions/list_or_dict"},
                 "network": {"type": "string"},
                 "pull": {"type": "boolean"},
                 "target": {"type": "string"},


ndeloof (approved):

build.md
```diff
@@ -227,6 +227,28 @@ Cache target is defined using the same `type=TYPE[,KEY=VALUE]` syntax defined by
...
+```
+
+```yml
+build:
+  context: .
+  additional_contexts:
+    project: /path/to/context
+```
+
+Illustrative examples of how this is used in BuildX can be found [here](https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#-additional-build-contexts---build-context).
```
milas, line 250:
```suggestion
Illustrative examples of how this is used in Buildx can be found [here](https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#-additional-build-contexts---build-context).
```

The most important, obviously

build.md
```diff
@@ -227,6 +227,28 @@ Cache target is defined using the same `type=TYPE[,KEY=VALUE]` syntax defined by
 
 Unsupported cache target MUST be ignored and not prevent user from building image.
 
+### additional_contexts
+
+`additional_contexts` defines a list of named contexts the image builder SHOULD use during image build.
```
milas, line 232: Should we specify the behavior if the builder doesn't support additional contexts? Warning or error? 🤔

build.md
```diff
@@ -227,6 +227,28 @@ Cache target is defined using the same `type=TYPE[,KEY=VALUE]` syntax defined by
...
+
+`additional_contexts` can be a mapping or a list:
+
+```yml
+build:
+  context: .
+  additional_contexts:
+    - project=/path/to/context
```
milas, line 240: Might be a good idea to include a couple examples of other "types", i.e. Git URLs, `docker-image://`

We could even specifically call out that the values are strings but validation beyond that is builder-specific (e.g. even within BuildKit, it'll depend on the version whether it supports it)

build.md
```diff
@@ -227,6 +227,28 @@ Cache target is defined using the same `type=TYPE[,KEY=VALUE]` syntax defined by
 
 Unsupported cache target MUST be ignored and not prevent user from building image.
 
+### additional_contexts
```
milas, line 230: In `docker-bake.hcl`, this is `contexts`. I wonder if we should try to remain consistent here (though having both `context` & `contexts` is 😬)

milas (commented):
This makes sense to me! Interested to hear thoughts from @tonistiigi @crazy-max @jedevc

build.md
laurazard: In the linked issue me and @ndeloof discussed this. I was against keeping `contexts` because I think it's going to introduce confusion (context is mandatory, these are just additional contexts for the builder to use and is optional)

build.md
laurazard: Do we do this for other features of this type?

build.md
laurazard: 😅

build.md
laurazard: (I'm okay with `named_contexts` as well)

build.md
```diff
@@ -227,6 +227,34 @@ Cache target is defined using the same `type=TYPE[,KEY=VALUE]` syntax defined by
...
+
+`additional_contexts` defines a list of named contexts the image builder SHOULD use during image build.
```
crazy-max, line 232: I think we should be aligned with current syntax buildx to avoid confusion: https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#-additional-build-contexts---build-context

So I would say `build_contexts` would be better or even better be aligned with [bake specification](https://docs.docker.com/build/bake/file-definition/#specification) so it would be `contexts`.

laurazard: we had this discussion here too https://github.com/compose-spec/compose-spec/pull/307#discussion_r1110116993, and here https://github.com/compose-spec/compose-spec/issues/306. I'd be okay with `build_contexts` too (as was my initial proposal), but I feel like `contexts` is too similar to the already existing `context`. cc @ndeloof @milas, maybe we can try to get all the discussion in one place 😅

ndeloof: sounds weird to me to define `build_contexts` inside the `build` section.
the buildx docs mentioned ^ use `addtional-(build)-contextx`, so my suggestion

build.md
laurazard: Further discussion in https://github.com/compose-spec/compose-spec/pull/307#discussion_r1110358705

build.md
laurazard: fwiw `named_contexts` is also an option here

build.md
```diff
@@ -227,6 +227,34 @@ Cache target is defined using the same `type=TYPE[,KEY=VALUE]` syntax defined by
...
+```yml
+build:
+  context: .
+  additional_contexts:
+    resources: /path/to/resources
+    app: docker-image://my-app:latest
+    source: https://github.com/myuser/project.git
+```
+
+When used as a list, the syntax should follow the `NAME=VALUE` format, where `VALUE` is a string. Validation beyond that is the responsibility of the image builder (and builder specific).
```
EricHripko, line 254:
```suggestion
When used as a list, the syntax should follow the `NAME=VALUE` format, where `VALUE` is a string. Validation beyond that is the responsibility of the image builder (and is builder specific).
```
tiny addition

EricHripko (approved):
Sounds good ✅ Thank you for putting this together 🙌

build.md
laurazard: Nice catch!

build.md
glours: Agree with @ndeloof , we're already in a `build` section and the `context` attribute already defines the default context, so `build_contexts` is not the best option here. 
`additional_contexts` or `named_contexts` seem to be better suitable names, even if I prefer the 1st one  which is more meaningful

build.md
glours: [The spec's default rule](https://github.com/compose-spec/compose-spec/blob/4ccbdd71694d97c3b011daba6f1d713e745c7807/spec.md?plain=1#L23) lets the implementation decide what to do with optional attributes.
But we can add something like that
`The Compose implementation SHOULD warn that the builder does not support additional contexts and MAY list the unused contexts.`

glours (approved):
Sounds good to me, maybe adding a hint when the constructor doesn't support additional contexts could be a great addition

build.md
laurazard: Sounds good, thanks for the suggestion! I updated the description to include it.

build.md
laurazard: If most of us agree, maybe we can keep `additional_contexts` then :)

jedevc: I have a preference towards `contexts` for existing consistency with bake, but it seems as though that's not the direction this is headed. It feels frustrating from a user-experience though that we'd have different keys for different scenarios: the command line and build-push-action (`--build-context` and `build-contexts`), bake (`contexts`) and compose (`named_contexts`/`additional_contexts`). I'm tempted to propose consistency with command line options that users can already use today with buildx and other build tools like buildah - either `contexts` or `build-contexts` sound good to me for this. 

That said, I don't want to stall the discussion, I'm alright with `additional_contexts` and `build_contexts` if there's a strong majority here.

glours: @EricHripko any thoughts on this one?

EricHripko: Having `build.build_contexts` seems a bit repetitive - it's already under `build` section, so `build_` prefix seems excessive. If we're really keen to keep consistent naming, then I would suggest that `build.context` acquires a long(ish?) form syntax - something akin to:
```yaml
# Single context
context: .
context: path/to/dir

# Multiple contexts
context:
  - .
  - alpine=docker-image://alpine@sha256:0123456789
context:
  - path/to/dir
  - project=path/to/project/source
```
From user perspective, the split between primary and additional build contexts seems like a bit of implementation details leaking into user space. Essentially, I'd consider all inputs to my build to be my build context.

In my (rather anecdotal) experience, I've seen very few developers be familiar with the newer `docker buildx` CLI and even fewer then heard/used bake in any shape or form. From that perspective, I am more keen on Compose spec making sense/being user-friendly as I'd expect the majority of developers to exclusively reside in this world.

If we have to introduce a new attribute for this, then I guess I could live with `additional_contexts`.

laurazard: I agree re: buildx and bakefile syntax, generally consistency is good but in this case I'd prioritize readability/clarity of the spec vs. strict adherence to the other syntaxes.

Re: multiple contexts in the `context` key though, I hadn't though about that! I'm okay with it, I only wonder if it's a bit confusing having a list where one unnamed entry is required and the (optional) others must be named – but I'm open to it if others agree with it.

ndeloof: I agree. This would be fine if we didn't had to manage this distinction between the (required) main context and the named ones. But here I wonder this will bring some confusion, so I prefer we introduce an explicit `named_contexts` field

EricHripko: Does the default/required context also have a name? BuildKit docs seem to suggest it's named `context`: `buildctl build --frontend=dockerfile.v0 --local context=. --local dockerfile=.`

PR #308, opened 2023-02-17
split compose spec.md

aevesdocker:

The spec.md file is incredibly long which makes it very hard to navigate (unless you love your scroll wheel or using CTRL+F). 

Links to https://github.com/docker/docs/pull/16740


laurazard:
I do enjoy opening 1 (actually 2 since we have `spec.md` and `build.md`) and just searching for what I want with CTRL-F 😅 

That said, I'm not strictly against the changes. I'm curious what other's thoughts are on this, and also on what other similar projects usually do (iirc these types of documents tend to be one long one vs. split out over a lot of files).

Other than that, I do wonder if the proposed split is the best – I think completely transversal functionalities can make sense to split out (like `interpolation.md` or `extension.md`), but wonder if things like `status.md` should be separate from `overview.md`.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,20 @@
...
+      -
+        name: Checkout
+        uses: actions/checkout@v3
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      -
+        name: Run
+        run: |
+          make validate-spec
```
laurazard, line 20: nit: add a newline at the end here.

ndeloof (approved):
:+1 on splitting this loooong file into subsections, especially if this helps keeping the docker docs aligned with the compose-spec for users to get relevant information

EricHripko:
I don't mind the split so long as places where we present the spec to users are easily searchable.

laurazard (approved):
LGTM

PR #309, opened 2023-02-17
cmd: update schema to account for empty overrides

milas:
* Clarify behavior around ignoring/clearing a base entrypoint or command using an empty list (`[]`) or emtpy string (`''`)
* Clarify handling of explicit `null` in commands
* Update schema to allow explicit `null` & use common type

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -429,17 +429,21 @@ cgroup_parent: m-executor-abcd
 
 ### command
 
-`command` overrides the default command declared by the container image (i.e. by Dockerfile's `CMD`).
+`command` overrides the default command declared by the image (i.e. by Dockerfile's `CMD`).
 
 ```
 command: bundle exec thin -p 3000
 ```
 
-The command can also be a list, in a manner similar to [Dockerfile](https://docs.docker.com/engine/reference/builder/#cmd):
+The value can also be a list, in a manner similar to [Dockerfile](https://docs.docker.com/engine/reference/builder/#cmd):
 
 ```
 command: [ "bundle", "exec", "thin", "-p", "3000" ]
 ```
+
+If the value is `null`, the default command from the image MUST be used.
+
+If the value is `[]` (emtpy list) or `''` (empty string), the default command declared by the image MUST be ignored, i.e. overridden to be empty.
 
 ### configs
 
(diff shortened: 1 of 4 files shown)


laurazard (approved):

spec.md
```diff
@@ -415,18 +415,22 @@ cgroup_parent: m-executor-abcd
 
 ### command
 
-`command` overrides the default command declared by the container image (i.e. by Dockerfile's `CMD`).
+`command` overrides the default command declared by the image (i.e. by Dockerfile's `CMD`).
```
EricHripko, line 418: Let's not remove the `container` image qualifier unless it's inaccurate

spec.md
```diff
@@ -415,18 +415,22 @@ cgroup_parent: m-executor-abcd
...
-The command can also be a list, in a manner similar to [Dockerfile](https://docs.docker.com/engine/reference/builder/#cmd):
+The value can also be a list, in a manner similar to [Dockerfile](https://docs.docker.com/engine/reference/builder/#cmd):
 
 ```
 command: [ "bundle", "exec", "thin", "-p", "3000" ]
 ```
 
+If the value is `null`, the default command from the image MUST be used.
+
+If the value is `[]` (emtpy list) or `''` (empty string), the default command declared by the image MUST be ignored, i.e. overridden to be empty.
```
EricHripko, line 432:
```suggestion
If the value is `[]` (empty list) or `''` (empty string), the default command declared by the image MUST be ignored, i.e. overridden to be empty.
```
typo

spec.md
```diff
@@ -738,6 +745,10 @@ entrypoint:
   - vendor/bin/phpunit
 ```
 
+If the value is `null`, the default entrypoint from the image MUST be used.
+
+If the value is `[]` (emtpy list) or `''` (empty string), the default entrypoint declared by the image MUST be ignored, i.e. overridden to be empty.
```
EricHripko, line 750:
```suggestion
If the value is `[]` (empty list) or `''` (empty string), the default entrypoint declared by the image MUST be ignored, i.e. overridden to be empty.
```
typo

EricHripko (commented):
Thank you for contributing to the spec! I've left a few comments inline 💬

spec.md
milas: Oops, good catch. Restored in c1d3e328a110d5a238bbb69cf310d67d7b3cc078

milas:
@laurazard Thanks for the ping!

@EricHripko Thanks for the review, apologies for missing it previously!

This should be good to go now - I accepted the code review feedback and copied the `spec.md` changes to `05-services.md` and re-generated 😅

ndeloof (approved):

glours (approved):
LGTM

Issue #298, opened 2022-12-24
inline Dockerfile


ndeloof:
**What is the problem you're trying to solve**
A common use-case is to customize an existing image with some minor changes. This requires a minimal Dockerfile with only a few lines. It would be nice one can declare such a Dockerfile directly inside compose.yaml, inlined

**Describe the solution you'd like**

Use yaml multiline literals with the pipe operator
We can introduce a new dedicated attrivube:
```yaml
services:
   demo:
     build:
        dockerfile-content: |
           FROM alpine
           RUN apk --no-cache add curl
```
we also can introduce a custom notation to declare inline Dockerfile
maybe we can use a custom yaml type `!inline` to mark such a block
```yaml
services:
   demo:
     build:
        dockerfile: !inline |
           FROM alpine
           RUN apk --no-cache add curl
```

**Additional context**
https://github.com/docker/compose/issues/8077

laurazard:
This is really cool! I'd lean towards custom yaml type over adding a new attribute, but I don't have super strong feelings one way or another.

ndeloof:
With some experiment, I noticed it's not trivial in the parser to manage proposed syntax:
```yaml
dockerfile: !inline | ... 
```
as the tag applies to the value, not the key. Would need to write:
```yaml
!inline dockerfile: | ... 
```
which sounds weird to me.

So I ended by adopting bake's `dockerfile-inline` additional key, mutually exclusive with `dockerfile`.

thaJeztah:
I'm wondering if this will be confusing for users w.r.t. interpolating environment variables / build-args. I know there's already a lot of confusion "who's responsible" for interpolating with combinations of `env`, `env_file`, and `.env` (as well as environment variables from the current environment). Adding the Dockerfile into the compose file itself may make this even more confusing.


```yaml
services:
   demo:
     env:
       - FOO=baz       
     env_file:
       - ./a.env
     build:
        dockerfile-content: |
          # syntax=docker/dockerfile:1
          
          ARG FOO=bar
           
          FROM alpine:${FOO}
          RUN apk --no-cache add curl
          COPY <<-"eot" /app/script.sh
             echo hello ${FOO}
          eot
          RUN <<-'eot'
             echo hello ${FOO}
          eot
          RUN echo ${FOO}
        args:
          FOO: hello
          BAR: ${FOO}
```

ndeloof:
[...]

PR #315, opened 2023-03-02
introduce dockerfile-inline

ndeloof:
define `dockerfile_inline` for user to declare a (simple) Dockerfile as inlined content directly in a compose-file.
attribute named has been chosen to align with bake's dockerfile-inline https://docs.docker.com/build/bake/file-definition/ but using underscore separator for consistency with other attributes

Fixes https://github.com/compose-spec/compose-spec/issues/298

### DIFF
--- a/build.md
+++ b/build.md
@@ -112,13 +112,30 @@ build:
 
 `dockerfile` allows to set an alternate Dockerfile. A relative path MUST be resolved from the build context.
 Compose implementations MUST warn user about absolute path used to define Dockerfile as those prevent Compose file
-from being portable.
+from being portable. When set, `dockerfile_inline` attribute is not allowed and a Compose Implementation SHOULD 
+reject any Compose file having both set.
 
 ```yml
 build:
   context: .
   dockerfile: webapp.Dockerfile
 ```
+
+### dockerfile_inline
+
+`dockerfile_inline` allows to define Dockerfile content as inlined string in a Compose file. When set, `dockerfile` 
+attribute is not allowed  and a Compose Implementation SHOULD reject any Compose file having both set.
+
+Use of YAML multi-line string syntax is recommended to define Dockerfile content:
+
+```yml
+build:
+  context: .
+  dockerfile_inline: |
+    FROM baseimage
+    RUN some command
+```
+
 
 ### args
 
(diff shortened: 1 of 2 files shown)


build.md
```diff
@@ -112,14 +112,27 @@ build:
...
 
 ```yml
 build:
   context: .
   dockerfile: webapp.Dockerfile
 ```
 
+### dockerfile_inline
+
+`dockerfile_inline` allows to define Dockerfile content as inlined string in a compose.yaml file. When set, `dockerfile` attribute is not allowed. Yaml multi-line string syntax is recommended to define Dockerfile content
```
EricHripko, line 125:
```suggestion
`dockerfile_inline` allows to define Dockerfile content as inlined string in a Compose file. When set, `dockerfile` attribute is not allowed. YAML multi-line string syntax is recommended to define Dockerfile content
```
Minor rewording suggestions

EricHripko:
The proposal makes sense - I was wondering if we could let `dockerfile` attribute be either file contents or file path, but I guess this would lead to peculiar edge cases and/or failure modes 😬

ndeloof:
> I was wondering if we could let dockerfile attribute be either file contents or file path, but I guess this would lead to peculiar edge cases and/or failure modes 😬

Yes, I expect this could bring some weird issues, so better use an explicit attribute. We could later investigate introducing more magic

build.md
```diff
@@ -112,14 +112,27 @@ build:
 
 `dockerfile` allows to set an alternate Dockerfile. A relative path MUST be resolved from the build context.
 Compose implementations MUST warn user about absolute path used to define Dockerfile as those prevent Compose file
-from being portable.
+from being portable. When set, `dockerfile_inline` attribute is not allowed.
```
glours, line 115: Maybe we should be clear on the behaviour expected when this case will happen? Should the Compose runtime refuse the Compose file as invalid or warn the user that the `dockerfile` will be used instead of the `inline` definition?

laurazard: I read it before as an implementation should refuse a Compose file that has both (since it's not allowed), but we could make that clearer 👍

laurazard (approved):
A bit sad about not having the custom yaml type, but it's probably for the best 😞

LGTM ✅

glours (approved):
LGTM

PR #316, opened 2023-03-04
Add podman-compose to README

bluecmd:
I just learned about podman-compose, and I had previously dismissed podman by default as it didn't have a compose alternative listed in this README.

N/A


EricHripko (approved):
Hi @bluecmd 👋 Thank you for contributing to the spec - looks good to me ✅

glours (approved):
No update since last October 🤔 
Anyway as we don't have a full conformance tests suite there is no blocker to add Podman Compose as a specification implementation

Issue #73, opened 2020-05-06
Define remote services and depend on them using healthchecks


k-miras:
**What is the problem you're trying to solve**
Docker-compose file services depends on services on another machine and we need to run our service only if services on remote machine are up.

**Describe the solution you'd like**
To use `depends_on` to run services on current machine after services on another machine that described as a `service` in `docker-compose.yml` file on current machine. Use `healthcheck` to verify whether remote service is up or down

**Additional context**
That's very basic way to solve the same problems in easier way that are solved using other types of container orchestration

EricHripko:
Sounds like your use case may be satisfied by `condition: service_healthy`, which I'm trying to bring in from v2 Compose in https://github.com/compose-spec/compose-spec/issues/68 and https://github.com/compose-spec/compose-spec/pull/72.

k-miras:
@EricHripko it could be partially satisfied, but there are some points that PR doesn't cover - defining remote service means that in volume mounts, networking setting, hardware limitations and lots the other constraints wouldn't work. In this Issue I'm trying to work with remote service as a black box and everything that I know about the black box is how to check is it alive using `healthcheck`. So, PR for this issue needs proper design

acran:
How about using a local service to verify the healthiness of the remote service like this:
~~~yml
services:
  app:
    depends_on:
      remote_service_monitor:
        condition: service_healthy
  remote_service_monitor:
    # ...
~~~

where `remote_service_monitor` is a simple _local_ container which just monitors the _remote_ service and exposes the results in its own health status. A very simple example to verify a remote web service is up could be:
~~~yml
services:
  remote_service_monitor:
    image: alpine
    # just to keep the container alive
    command: sh -c 'while true; do sleep 999; done'
    healthcheck:
      test: "wget -qO - https://remote.example.com/healthcheck"
~~~

ndeloof:
We used to define such "remote service" as "external services" for internal discussion. This is something that we need to address, maybe we to follow @acran suggestion an apply same mechanism used for existing volumes/network/secrets we want to use without compose to manage those:

```yaml
services:
  app:
    depends_on:
      remote_service:
        condition: service_healthy

  remote_service:
    external: true
    # need some way to define how we resolve this service
    healthcheck:
      url: https://remote.example.com/healthcheck
```

[...]

PR #317, opened 2023-03-07
introduce service.external

ndeloof:
introduce `external` service attribute to declare a service as "managed outside compose lifecycle"

See https://github.com/compose-spec/compose-spec/issues/73

### DIFF
--- a/05-services.md
+++ b/05-services.md
@@ -902,6 +902,40 @@ duplicates resulting from the merge are not removed.
 
 Any other allowed keys in the service definition should be treated as scalars.
 
+## external
+
+[![Compose NEXT RELEASE](https://img.shields.io/badge/compose-NEXT-blue?style=flat-square)](https://github.com/docker/compose/releases/NEXT)
+
+A service can be declared with `external` attribute. Such a service is not directly managed by Compose. Compose delegates lifecycle
+to a third-party component identified by the `type` attribute. Comparable to a volume/plugin driver, `options` can be set to pass
+arguments to this third-party component, prefixed with uppercase service name.
+
+```yaml
+services:
+  database:
+     external: 
+       type: cloud-db
+       options:
+         credentials: ${API_TOKEN}
+```
+
+The services which declare a dependency to an `external` service get additional environment variables set, produced by the
+external service component. In this illustration example, `app` service will get `DATABASE_URL` and `DATABASE_PASSWORD` 
+environment variables injected, based on values produced by third-party `cloud-db` component. 
+
+```yaml
+services:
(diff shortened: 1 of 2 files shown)


glours:
This changes of this PR should be moved to the [05-services.md](https://github.com/compose-spec/compose-spec/blob/master/05-services.md) file

05-services.md
```diff
@@ -784,6 +784,33 @@ duplicates resulting from the merge are not removed.
...
+
+
+```yaml
+services:
+  app:
+    image: mycompay/app
+    environment:
+      - DB_URL=${DB_URL}
+    depends_on:
+      - db
```
nicks, line 809: in this example, how will compose verify that db is ready before it starts `app`? (the above says that we can't even assume that `db` is a container)

ndeloof: with this initial design, it just can't. It's user (or third party tool) responsibility to manage service lifecycle and enforce it's running and healthy.
I wonder how to define some `inspect`, `create` and `delete` logic to be applied to such a service, within the compose specification, offering the ability to run external tools.

ndeloof: My assumption here is that the spec hardly can define a portable way to manage lifecycle for a non-container based service (let's say, an AWS RDS database). A Compose Implementation dedicated to this use case could offer extensions to better support this usage, with additional attributes (maybe relying on `x-*` extensions)

nicks: if other containers can't interact with it, then why put `external: true` in the compose file at all? What problem does it solve then?

Or are you imagining that `external: true` is just there for documentation purposes, and you have to add more fields to it to add behavior?

ndeloof: > if other containers can't interact with it,

Obviously they can. They can connect to a RDS database or a FaaS function. But there's no generic way to define lifecycle management for such resources, especially considering compose (at least, Docker Compose) is a client-side implementation which might not have access to the same networks.

> What problem does it solve then? 

My compose file used to declare a database service and I have dependencies defined for this purpose. Then I want to switch to a RDS database. I would need to fully rewrite my compose file to "remove" this db service. With such an `external: true` I can provide an override to just disable service management by compose, and assume some third-party tool ran just before to enforce RDS database is up.

nicks: Aha! I missed the override use case. Knew I was missing something, thanks!

ndeloof:
updated to reflect https://github.com/compose-spec/compose-go/pull/758 cc @glours

ndeloof:
obsolete, see https://github.com/compose-spec/compose-spec/pull/594

PR #318, opened 2023-03-14
Simpler approach building spec.md

ndeloof:
For some weird reason I can't get `markdown-include` to be installed on my docker image
So I checked the spec.md build logic, and came with a simpler approach just appending files


spec.md
```diff
@@ -2540,19 +2538,20 @@ Interpolation can also be nested:
 Other extended shell-style features, such as `${VARIABLE/foo/bar}`, are not
 supported by the Compose specification.
 
-You can use a `$` (double-dollar sign) when your configuration needs a literal
-dollar sign. This also prevents Compose from interpolating a value, so a `$`
+You can use a `$$` (double-dollar sign) when your configuration needs a literal
```
milas, line 2541: Oh this is an interesting (albeit unintentional 😛) fix

milas (commented):
Not sure why the `validate` job is unhappy, but this LGTM!

Will approve once CI is green 🟢

spec.md
ndeloof: CI fails because of this!

PR #323, opened 2023-03-24
Make `network_mode` and `networks` mutually exclusive attributes

### LINKED ISSUE #322: Unpredictable outcomes when setting `network_mode` and `networks`
## Context

In the `service` level configuration, it's possible to configure both `network_mode` and `networks` attributes. The interaction between these attributes is not specified, and (as of today) produces unexpected behaviors in Docker Compose.

Possible combinations of these attributes are:

| network_mode | networks | expected result |
|----------------|----------|-----------------|
| host                  | -               | clear                   |
| host                  | yes           | ??                       |
| none                 | -               | clear                   |
| none                 | yes           | ??                       |
| service:[a]       | -               | clear                   |
| service:[a]       | yes           | ??                        |  

For a more concrete example, take the following Compose file:

```yaml
services:
  a:
    image: alpine
    command: ping c
    network_mode: service:b
    networks:
      - n
  b:
    image: alpine
    command: top
  c:
    image: alpine
    command: top
    networks:
      - n

networks:
  n:
```

Currently, Docker Compose does not reject this file. However, if one were to try to run this file, service `a` is not able to ping `c` due to it's configured `network_mode` (fails with `ping: bad address 'c'`). This is strange, unexpected behaviour – my containers are on the same network, why can't they talk to each other?

Other shortcomings are less convoluted but should still be addressed:
- setting `network_mode: host` and `networks` results, in Docker Compose, in an error being returned by the daemon (which makes sense, these aren't compatible options, but the resulting message can be confusing and we could address this from the start by telling the user that this configuration is invalid)
- setting `network_mode: none` and `networks` doesn't produce an error, but the `network_mode` is ignored and the container is granted networking anyway to the defined networks. These should also be incompatible options.

## Describe the solution you'd like
We could define multiple rules for how the different combinations of these options should interact (and I welcome discussion on that) but I think the clearest, most explicit way forward here is to define, in the Spec, that these are incompatible attributes, and implementations SHOULD reject configurations containing both.

I was going to start by just making a PR to the Spec to address this, but I thought explaining the need for this more clearly was a better 1st step (and maybe someone has better ideas on how to handle it).

laurazard:

Makes explicit that a service configuration can not contain both `network_mode` and `networks`.

Fixes #322


ndeloof (approved):

05-services.md
```diff
@@ -1060,6 +1060,9 @@ specification define specific values which MUST be implemented as described if s
     network_mode: "service:[service name]"
 ```
 
+When set, `networks` attribute is not allowed and a Compose Implementation SHOULD reject any
```
glours, line 1063:
```suggestion
When set, `networks` attribute is not allowed and a Compose implementation SHOULD reject any
```

spec.md
```diff
@@ -1311,6 +1311,9 @@ specification define specific values which MUST be implemented as described if s
     network_mode: "service:[service name]"
 ```
 
+When set, `networks` attribute is not allowed and a Compose Implementation SHOULD reject any
```
glours, line 1314:
```suggestion
When set, `networks` attribute is not allowed and a Compose implementation SHOULD reject any
```

glours (changes requested):
We should use `Compose implementation` keep the naming consistency.
I'll open a PR to fix the few other `Compose implementation` occurrences using capital `I`

glours (approved):
LGTM

Issue #324, opened 2023-03-27
A customer is reporting that an example is missing


voonyee:
[08-configs.md](https://docs.docker.com/compose/compose-file/08-configs/)
A customer reporting the below:

"
Looks like an example is missing on page https://docs.docker.com/compose/compose-file/08-configs/.
Seems there should be an example after "be obtained by lookup." words. Just letting you know. It would be great if you'll add it.
"
Case number: 00059552

PR #333, opened 2023-04-06
fix config spec

ndeloof:
Fixes https://github.com/compose-spec/compose-spec/issues/324


glours (approved):
LGTM

PR #339, opened 2023-04-17
clarifiy support for relative paths in additional contexts

ndeloof:

see https://github.com/compose-spec/compose-go/pull/385#issuecomment-1500505431 for context


build.md
```diff
@@ -284,7 +284,9 @@ build:
     source: https://github.com/myuser/project.git
 ```
 
-When used as a list, the syntax should follow the `NAME=VALUE` format, where `VALUE` is a string. Validation beyond that is the responsibility of the image builder (and is builder specific).
+When used as a list, the syntax should follow the `NAME=VALUE` format, where `VALUE` is a string. Validation beyond that is the responsibility of the image builder (and is builder specific). A Compose implementation MUST support at least
+passing a path to a directory (absolute or realite) and an git repository URL, like [context](#context) does. Other context flavours MUST be previxed to avoid
+ambiguity with a `type://` prefix.
```
laurazard, line 289:
```suggestion
When used as a list, the syntax should follow the `NAME=VALUE` format, where `VALUE` is a string. Validation beyond that is the responsibility of the image builder (and is builder specific). A Compose implementation MUST support
at least absolute and relative paths to a directory AND Git repository URLs, like [context](#context) does. Other context flavours MUST be prefixed to avoid
ambiguity with a `type://` prefix.
```

milas (approved):

PR #341, opened 2023-04-21
Deploy build status tidy

aevesdocker:

Received feedback from HotJar that deploy.md and build.md need a tidy up as the language is confusing. 

This PR also does a tidy up of the status.md page

### DIFF
--- a/01-status.md
+++ b/01-status.md
@@ -1,25 +1,25 @@
-## Status of this document
+# Status of this document
 
 This document specifies the Compose file format used to define multi-containers applications. Distribution of this document is unlimited.
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
-### Requirements and optional attributes
+## Requirements and optional attributes
 
 The Compose specification includes properties designed to target a local [OCI](https://opencontainers.org/) container runtime,
-exposing Linux kernel specific configuration options, but also some Windows container specific properties, as well as cloud platform features related to resource placement on a cluster, replicated application distribution and scalability.
+exposing Linux kernel specific configuration options, but also some Windows container specific properties. It is also designed for cloud platform features related to resource placement on a cluster, replicated application distribution, and scalability.
 
-We acknowledge that no Compose implementation is expected to support **all** attributes, and that support for some properties
-is Platform dependent and can only be confirmed at runtime. The definition of a versioned schema to control the supported
+We acknowledge that no Compose implementation is expected to support all attributes, and that support for some properties
+is platform dependent and can only be confirmed at runtime. The definition of a versioned schema to control the supported
 properties in a Compose file, established by the [docker-compose](https://github.com/docker/compose) tool where the Compose
 file format was designed, doesn't offer any guarantee to the end-user attributes will be actually implemented.
 
-The specification defines the expected configuration syntax and behavior, but - until noted - supporting any of those is OPTIONAL.
+The specification defines the expected configuration syntax and behavior. Unless noted, supporting any of these is OPTIONAL.
 
-A Compose implementation to parse a Compose file using unsupported attributes SHOULD warn user. We recommend implementors
+A Compose implementation to parse a Compose file using unsupported attributes SHOULD warn users. We recommend the following implementors
 to support those running modes:
 
-* default: warn user about unsupported attributes, but ignore them
-* strict: warn user about unsupported attributes and reject the compose file
-* loose: ignore unsupported attributes AND unknown attributes (that were not defined by the spec by the time implementation was created)
+* Default: warn the user about unsupported attributes, but ignore them
+* Strict: warn the user about unsupported attributes and reject the Compose file
+* Loose: ignore unsupported attributes AND unknown attributes (that were not defined by the spec by the time implementation was created)
 
--- a/01-status.md
+++ b/01-status.md
@@ -1,10 +1,10 @@
-# Status of this document
+## Status of this document
 
 This document specifies the Compose file format used to define multi-containers applications. Distribution of this document is unlimited.
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
-## Requirements and optional attributes
+### Requirements and optional attributes
 
 The Compose specification includes properties designed to target a local [OCI](https://opencontainers.org/) container runtime,
 exposing Linux kernel specific configuration options, but also some Windows container specific properties. It is also designed for cloud platform features related to resource placement on a cluster, replicated application distribution, and scalability.
--- a/build.md
+++ b/build.md
@@ -8,7 +8,7 @@
 
 ## Introduction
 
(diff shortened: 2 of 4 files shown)


ndeloof:
Thanks for your contribution. Not being a native english speaker I will let my colleagues review proposed changes

01-status.md
```diff
@@ -1,25 +1,25 @@
-## Status of this document
+# Status of this document
```
glours, line 1: only changing the level of this part of the specification will make that section the root of all the others, which isn't what we expect. If you want to make this change, you'll have to modify all the titles in the others files.

glours (changes requested):
The change on the status title will affect the presentation of the whole `spec.md` file

01-status.md
aevesdocker: Ah silly me, I forgot how this would appear on spec.md as a whole.. Reverting now!

aevesdocker, line 1:
```suggestion
## Status of this document
```

01-status.md
```diff
@@ -1,25 +1,25 @@
...
 
 This document specifies the Compose file format used to define multi-containers applications. Distribution of this document is unlimited.
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
-### Requirements and optional attributes
+## Requirements and optional attributes
```
aevesdocker, line 7:
```suggestion
### Requirements and optional attributes
```

build.md
```diff
@@ -1,59 +1,55 @@
...
 
 * ToC
 {:toc}
 
 ## Introduction
 
-Compose specification is a platform-neutral way to define multi-container applications. A Compose implementation
-focusing on development use-case to run application on local machine will obviously also support (re)building
-application from sources. The Compose Build specification allows to define the build process within a Compose file
+The Compose specification is a platform-neutral way to define multi-container applications. A Compose implementation which focuses on running an application on a local machine will obviously need to alo support (re)building
```
laurazard, line 11:
```suggestion
The Compose specification is a platform-neutral way to define multi-container applications. A Compose implementation which focuses on running an application on a local machine will obviously need to also support (re)building
```

build.md
```diff
@@ -1,59 +1,55 @@
...
 ## Definitions
 
-Compose Specification is extended to support an OPTIONAL `build` subsection on services. This section define the
-build requirements for service container image. Only a subset of Compose file services MAY define such a Build
-subsection, others being created based on `Image` attribute. When a Build subsection is present for a service, it
-is *valid* for a Compose file to miss an `Image` attribute for corresponding service, as Compose implementation
-can build image from source.
+Compose Specification can be extended to support an OPTIONAL `build` subsection on services. 
+
+This section defines the build requirements for a service container image. Only a subset of Compose file services MAY define such a `build`
```
laurazard, line 19: This sort of reads like not ALL services in a Compose file MAY define a build section, while what it's trying to say is that not all services MUST define a build section as far as I can tell. Maybe we can reword this?

build.md
```diff
@@ -1,59 +1,55 @@
...
 
 Build can be either specified as a single string defining a context path, or as a detailed build definition.
 
-In the former case, the whole path is used as a Docker context to execute a docker build, looking for a canonical
-`Dockerfile` at context root. Context path can be absolute or relative, and if so relative path MUST be resolved
-from Compose file parent folder. As an absolute path prevent the Compose file to be portable, Compose implementation
-SHOULD warn user accordingly.
+In the former case, the whole path is used as a Docker context to execute a Docker build, looking for a canonical
+`Dockerfile` at the root of the directory. The path can be absolute or relative. If it is relative, it MUST be resolved
+from the Compose file parent folder. If it is absolute, the path prevents the Compose file from being portable so the Compose implementation SHOULD warn user accordingly.
```
laurazard, line 28: For consistency with previous wording changes, maybe use
```suggestion
from the Compose file parent folder. If it is absolute, the path prevents the Compose file from being portable so the Compose implementation SHOULD warn users accordingly.
```

build.md
```diff
@@ -1,59 +1,55 @@
...
 
-In the later case, build arguments can be specified, including an alternate `Dockerfile` location. This one can be
-absolute or relative path. If Dockerfile path is relative, it MUST be resolved from context path.  As an absolute
-path prevent the Compose file to be portable, Compose implementation SHOULD warn user if an absolute alternate
-Dockerfile path is used.
+In the later case, build arguments can be specified, including an alternate `Dockerfile` location. The path can be absolute or relative. If it is relative, it MUST be resolved
+from the Compose file parent folder. If it is absolute, the path prevents the Compose file from being portable so the Compose implementation SHOULD warn user accordingly.
```
laurazard, line 31: same here:
```suggestion
from the Compose file parent folder. If it is absolute, the path prevents the Compose file from being portable so the Compose implementation SHOULD warn user saccordingly.
```

build.md
```diff
@@ -1,59 +1,55 @@
...
+if the image was not found on registry. The Compose implementation MAY offer options to customize this behaviour by user
 request.
 
 ## Publishing built images
 
-Compose implementation with Build support SHOULD offer an option to push built images to a registry. Doing so, it
-MUST NOT try to push service images without an `Image` attribute. Compose implementation SHOULD warn user about
-missing `Image` attribute which prevent image being pushed.
+The Compose implementation with Build support SHOULD offer an option to push built images to a registry. When doing so, it
+MUST NOT try to push service images without an `image` attribute. The Compose implementation SHOULD warn user about the missing `Image` attribute which prevent images being pushed.
```
laurazard, line 44:
```suggestion
MUST NOT try to push service images without an `image` attribute. The Compose implementation SHOULD warn users about the missing `Image` attribute which prevent images being pushed.
```

build.md
```diff
@@ -126,9 +122,11 @@ services:
 
 ### dockerfile
 
-`dockerfile` allows to set an alternate Dockerfile. A relative path MUST be resolved from the build context.
-Compose implementations MUST warn user about absolute path used to define Dockerfile as those prevent Compose file
-from being portable. When set, `dockerfile_inline` attribute is not allowed and a Compose Implementation SHOULD 
+`dockerfile` lets you set an alternate Dockerfile. A relative path MUST be resolved from the build context.
```
laurazard, line 125:
```suggestion
`dockerfile` lets users set an alternate Dockerfile. A relative path MUST be resolved from the build context.
```

laurazard: or alternatively
```suggestion
`dockerfile` sets an alternate Dockerfile. A relative path MUST be resolved from the build context.
```

build.md
```diff
@@ -139,8 +137,8 @@ build:
 
 ### dockerfile_inline
 
-`dockerfile_inline` allows to define Dockerfile content as inlined string in a Compose file. When set, `dockerfile` 
-attribute is not allowed  and a Compose Implementation SHOULD reject any Compose file having both set.
+`dockerfile_inline` lets you define Dockerfile content as inlined string in a Compose file. When set, the `dockerfile`
```
laurazard, line 140:
```suggestion
`dockerfile_inline` lets users define Dockerfile content as inlined string in a Compose file. When set, the `dockerfile` 
```

laurazard: or 
```suggestion
`dockerfile_inline` defines Dockerfile content as inlined string in a Compose file. When set, the `dockerfile` 
```

laurazard (approved):
LGTM, just some little nits :) This is much nicer, thank you! 🥳

aevesdocker:
Thank you for the review @laurazard!

laurazard:
@milas @glours wanna TAL? 👀

glours (approved):

PR #343, opened 2023-04-26
Improved diagram

Omoba5:
The initial illustrative example was not as easy to read and not as elaborate

I created a better version of the sketch in the illustrative example without tampering with its meaning. I used mermaid markdown because I was advised to give great priority to Maintainability. 
Now the illustration is beautiful, maintainable and still being meaningful.

### DIFF
--- a/02-model.md
+++ b/02-model.md
@@ -34,20 +34,39 @@ The backend stores data in a persistent volume.
 
 Both services communicate with each other on an isolated back-tier network, while frontend is also connected to a front-tier network and exposes port 443 for external usage.
 
-```
-(External user) --> 443 [frontend network]
-                            |
-                  +--------------------+
-                  |  frontend service  |...ro...<HTTP configuration>
-                  |      "webapp"      |...ro...<server certificate> #secured
-                  +--------------------+
-                            |
-                        [backend network]
-                            |
-                  +--------------------+
-                  |  backend service   |  r+w   ___________________
-                  |     "database"     |=======( persistent volume )
-                  +--------------------+        \_________________/
+```mermaid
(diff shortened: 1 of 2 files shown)


ndeloof:
you have to run `make spec` to generate the one-pager spec file and add it to your commit

Omoba5:
I have done that now.

glours:
I don't know why but looking at the changes only show a modification which may not link to the purpose of this Pull Request

Omoba5:
It's because there was no file change in the last commit. I just ran `make spec` as advised. Please check the series of commits to see the changes. Or do you want me to rebase and then push as 2 commits?

Omoba5:
Oh, I just saw what you are talking about. the last commit reverted the real reason for the PR. Will rebasing and pushing all together fix this?

ndeloof:
the issue is you made change on `spec.md`, which is not the source of truth. You should have edited `02-model.md` and ran `make spec` to produce the one-pager.
Also, feel free to squash your commits so we don't keep all the back and forth after merge

glours:
@Omoba5 unfortunately no, the changes without selecting a specific commit should show all the changes from all commits, and here I see only 1 change unrelated with the diagram modification.
The issue here, is that your changes should be done in the `https://github.com/compose-spec/compose-spec/blob/master/02-model.md` file, the `spec.md` is generated and should not be modified directly

Omoba5:
Okay, I will implement that. Thank you for being patient with me.

Omoba5:
Done! This has to be it.

ndeloof:
Merged, thanks a lot for your contribution

PR #342, opened 2023-04-26
improved illustrative example in spec.md file Signed-off-by: Prince Ogabi ogabiprince@gmail.com

Omoba5:
**What this PR does / why we need it**:
The initial illustrative example was not as easy to read and not as elaborate

**What is the problem you're trying to solve**
I created a better version of the sketch in the illustrative example without tampering with its meaning.

Signed-off-by: Prince Ogabi <ogabiprince@gmail.com>

PR #344, opened 2023-04-27
convert diagram to mermaid

ndeloof:
use mermaid to define the model diagram, which renders as a (reasonably) nice image on github, but still is easy to compare on pull request for maintenance - actually way easier to edit and compare than ASCII-art based diagram
Can be edited using https://mermaid.live/ for those not comfortable with the syntax

see result on https://github.com/ndeloof/compose-spec/blob/mermaid/02-model.md

As suggested by discussions on https://github.com/compose-spec/compose-spec/pull/342


ndeloof:
closed in favor for https://github.com/compose-spec/compose-spec/pull/343

PR #351, opened 2023-05-11
fixing format of codeblock

dozjul:
in the rest of the file the codeblocks are seperated by the rest of the text by a blank line. As is usually best practice. However in this section it wasn't seperated by a blank line.

I changed the format by inserting a blank line to match the rest of the file

On https://docs.docker.com/compose/compose-file/build/ the code-block was rendered incorrectly, but only at this one code-block.  

---

A screenshot of the wrongly rendered section is attached below:

image

---

I assume, that it might relate to this code-block not being seperated from the text by a blank line.
And even if not, approving this merge request might improve readability for those who might read this in a txt-Editor, and improve the consistency of this file.
However, if this behaviour is intended feel free to reject this merge request.

**Which issue(s) this PR fixes**: I didn't find an issue relating to this. As this is an minor change, it should not need a issue associated (as stated by contributing guidelines).

### DIFF
--- a/build.md
+++ b/build.md
@@ -490,6 +490,7 @@ platform, otherwise users won't be able to run images they built.
 Compose implementations SHOULD report an error in the following cases:
 * when the list contains multiple platforms but the implementation is incapable of storing multi-platform images
 * when the list contains an unsupported platform
+
 ```yml
 build:
   context: "."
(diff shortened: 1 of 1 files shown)


laurazard (approved):
LGTM, thanks 🥳

ndeloof (approved):

PR #353, opened 2023-05-16
add Windows drive prefix to temp directory path to be sure Makefile works with Git Bash

glours:
Add Windows drive prefix to temp directory when executing a shell command like `cp` or `rm`

At least Makefile will work with Git Bash


aevesdocker (approved):
This fix worked for me!

milas (approved):

PR #356, opened 2023-05-16
add more info to fragements and extension

aevesdocker:

Fixes #


milas (approved):

10-fragments.md
```diff
@@ -33,3 +63,22 @@ volumes:
...
+      - CONFIG_KEY
+      - EXAMPLE_KEY
+      - DEMO_VAR
+  second:
+    image: another-image:latest
+    environment:
+      <<: *env
+      - AN_EXTRA_KEY
+      - SECOND_SPECIFIC_KEY
+```
```
ndeloof, line 82: this doesn't work. yaml merge `<<: *xx` can only be used with a mapping.

ndeloof (changes requested):
way better than previous doc
but example merging with a sequence is invalid

aevesdocker:
Thanks for catching that @ndeloof. I've updated the example with bits from your closed PR

aevesdocker:
Unsure why I'm suddenly getting build errors again and spec.md file appears deleted 🤔

ndeloof (approved):

PR #360, opened 2023-05-19
Engdocs 1283

aevesdocker:

This PR aligns the secrets.md file to our Style Guide. It also adds more headers to make the examples easier to skim over.


09-secrets.md
```diff
@@ -56,4 +62,7 @@ not managed by compose lifecycle, Compose implementations SHOULD reject the Comp
```
aevesdocker, line 62: I'm not sure what is meant on line 60 - would be good to make this clearer. @milas could you clarify for me?

09-secrets.md
```diff
@@ -1,28 +1,33 @@
 ## Secrets top-level element
 
-Secrets are a flavour of Configs focusing on sensitive data, with specific constraint for this usage. As the platform implementation may significantly differ from Configs, dedicated Secrets section allows to configure the related resources.
+Docker Compose provides a way for you to use secrets without having to use environment variables to store information. If you’re injecting passwords and API keys as environment variables, you risk unintentional information exposure. Environment variables are often available to all processes, and it can be difficult to track access. They can also be printed in logs when debugging errors without your knowledge. Using secrets mitigates these risks.
```
ndeloof, line 3: this is compose specification, not Docker Compose documentation. Should rephrase as `a Compose implementation` or similar

09-secrets.md
ndeloof: means having `external: true` makes attributes other than `name` irrelevant and should trigger an error

09-secrets.md
aevesdocker: I understand, I've changed it to just 'Compose' - very broad and general - as people reading this on docs.docker.com will be thrown with 'A Compose implementation' for example

09-secrets.md
aevesdocker: Thanks Nicolas, have made this clearer

09-secrets.md
```diff
@@ -1,28 +1,33 @@
 ## Secrets top-level element
 
-Secrets are a flavour of Configs focusing on sensitive data, with specific constraint for this usage. As the platform implementation may significantly differ from Configs, dedicated Secrets section allows to configure the related resources.
+Compose provides a way for you to use secrets without having to use environment variables to store information. If you’re injecting passwords and API keys as environment variables, you risk unintentional information exposure. Environment variables are often available to all processes, and it can be difficult to track access. They can also be printed in logs when debugging errors without your knowledge. Using secrets mitigates these risks.
 
-The top-level `secrets` declaration defines or references sensitive data that can be granted to the services in this
-application. The source of the secret is either `file` or `external`.
+Secrets are also useful for locally replicating a remote development setup with Kubernetes. Using secrets makes your application scalable.
```
ndeloof, line 5: how do compose secrets help "replicating  a remote development setup with Kubernetes" ?
AFAICT kube secrets are not exact equivalent for docker swarm nor compose secrets

Also, "Using secrets makes your application scalable" doesn't make sense to me. Scalability isn't involved here

09-secrets.md
```diff
@@ -1,28 +1,33 @@
...
+
+The top-level `secrets` element defines or references sensitive data that can be granted to your services. The source of the secret is set with either `file` or `external`.
```
ndeloof, line 7: The source can be `file` or `environment`. `external` is not a secret source, but a way to flag a resource as "already existing"

09-secrets.md
glours, line 5: I'm wondering if this type of content isn't more documentation one than specification one?
What we need on the specification level is an explanation of what a secret is, when or why using it. The current definition don't reach this goal, it's the same with this proposal, maybe we should take inspiration from [the Volumes part](https://github.com/compose-spec/compose-spec/blob/master/07-volumes.md) of the spec to provide a clearer definition of Secrets (and Configs?)

09-secrets.md
```diff
@@ -40,8 +47,7 @@ secrets:
 ```
 
 External secrets lookup can also use a distinct key by specifying a `name`. The following
-example modifies the previous one to look up for secret using a parameter `CERTIFICATE_KEY`. The
-the actual lookup key will be set at deployment time by [interpolation](12-interpolation.md) of
+example extends the previous example and has Compose looking for a secret using the parameter `CERTIFICATE_KEY`. The the actual lookup key is set at deployment time by [interpolation](12-interpolation.md) of
```
ndeloof, line 50: "The the"

09-secrets.md
```diff
@@ -51,9 +57,11 @@ secrets:
     name: "${CERTIFICATE_KEY}"
 ```
 
-If `external` is set to `true` and secret configuration has other but `name` attributes set, considering resource is
-not managed by compose lifecycle, Compose implementations SHOULD reject the Compose file as invalid.
+If `external` is set to `true`, all other attributes apart from `name` are irrelevant. If Compose detects any other attribute, it SHOULD reject the Compose file as invalid.
```
ndeloof, line 60: "the compose file" (lowercase)

ndeloof (changes requested):
this part of the spec deserves some improvements, but the general description is incorrect.

09-secrets.md
aevesdocker: From my conversation with Milas, my understanding was that these were two additional benefits of using Secrets. But perhaps I misunderstood. Is there something else I can put here? What would you recommend? Or is it as simple as Secrets are superior to environment variables when storing/using sensitive information?

ndeloof: There's no other benefits that comes to my mind

ndeloof (approved):

09-secrets.md
laurazard: Right. I think the right move here would be to have actual Compose docs, and have the Spec just be a Spec, otherwise concerns are mixed and the water gets a bit murky. There's a clear contrast between language elsewhere which just describes how a Compose file should be parsed and what a Compose implementation should do with that information vs this kind of language, and I don't think that information should reside within the Spec.

09-secrets.md
```diff
@@ -1,28 +1,31 @@
 ## Secrets top-level element
 
-Secrets are a flavour of Configs focusing on sensitive data, with specific constraint for this usage. As the platform implementation may significantly differ from Configs, dedicated Secrets section allows to configure the related resources.
+Secrets provides a more secure way of getting sensitive information in to your application's services, so you don't have to rely on using environment variables. If you’re injecting passwords and API keys as environment variables, you risk unintentional information exposure. Environment variables are often available to all processes, and it can be difficult to track access. They can also be printed in logs when debugging errors without your knowledge. Using secrets mitigates these risks.
```
laurazard, line 3: The tone/wording here contrasts with the rest of the specification (`if you're`, `so you don't`). The rest of the Specification generally doesn't address the reader directly, and uses a more formal tone. I'm still not convinced this information should reside here, but if we want to include it I'd rather it be changed to match the rest of the Specification.

laurazard (changes requested):

aevesdocker:
As per slack thread, will pick this work back up again once we're all aligned on the goal and wording of the Spec

PR #363, opened 2023-06-08
introduce `include` to depend on another compose file

ndeloof:
introduce `include` to depend on another compose file
this will allow to modularize complex compose file, share some common definition with other teams, and also resolve issue for users to extend a compose file from a distinct folder

illustrated by https://github.com/compose-spec/compose-go/pull/416
closes https://github.com/docker/compose/issues/318 after ... 9 years ! 😅

### DIFF
--- a/14-include.md
+++ b/14-include.md
@@ -1,0 +1,90 @@
+## Dependency on other compose projects
+
+A Compose application can declare dependency on another compose application, managed by another team
+or shared with others. This allows to keep a compose file reasonably complicated for the limited
+amount of resources a single team has to declare for it's own sub-domain within a larger deployment.
+
+`include` top-level section is used to define dependency to another compose application (or subdomain).
+Each path listed in this section must be loaded as an individual compose model with it's own project
+directory to resolve relative paths. 
+
+Once the application to be included has been loaded, all resources definitions are copied into the 
+current compose application model. Compose must warn user if some resource name conflict, and not 
+try to merge those. To enforce this, `include` must be evaluated after the compose file(s) selected 
+by user to define the Compose application model have been parsed and merged, so that conflicts 
+between included compose files and those selected are detected.
+
+`include` applies recursively: an included compose file which declares it's own `include` section
+will trigger those other files to be included as well.
+
+The resulting resources can be used in the including compose model for cross-services references.
+
+### short syntax
+
+Short syntax only defines path to another compose file. File is loaded with parent
+folder as project directory, and optional `.env` file being loaded to define variables default values
+for interpolation, while local project environment can override those values. 
+
+```yaml
+include:
+  - ../commons/compose.yaml
+  - ../another_domain/compose.yaml
+
+services:
+  webapp:
+    depends_on:
+      - included-service # defined by another_domain
+```
+
+In this illustration example, when loading compose file, both `../commons/compose.yaml` and 
+`../another_domain/compose.yaml` are loaded as individual compose projects. Relative paths 
+in compose files being refered by `include` are resolved relative to their own compose 
+file path, not based on local project directory. Variables are interpolated using values set in
+`.env` optional file in same folder, and can be overriden by local project environment.
+
+### long syntax
+
+Long syntax offer fine-grain control over the sub-project parsing:
+
+```yaml
+include:
+   - path: ../commons/compose.yaml
+     project_directory: ..
+     env_file: ../another/.env
+```
+
+#### path
+`path` is required and defines the location of the compose file(s) to be parsed and included into
+local compose model. `path` can be set either to a string when a single compose file is involved,
+or to a list of strings when multiple compose files need to be [merged together](14-merge.md) to 
+define the compose model to be included in local application.
+
+```yaml
+include:
+   - path: 
+       - ../commons/compose.yaml
+       - ./commons-override.yaml
+```
+
+#### project_directory
+`project_directory` defines base path to resolve relative paths set in compose file. It defaults to 
+the directory of the included compose file.
+
+#### env_file
+`env_file` defines an environment file(s) to use to define default values when interpolating variables
+in the compose file being parsed. It defaults to `.env` file in the `project_directory` for the compose 
+file being parsed. 
+
+`env_file` can be set either to a string or a list of strings when multiple env_file need to be merged
+to define project environment.
+
+```yaml
+include:
+   - path: ../another/compose.yaml
+     env_file:
+       - ../another/.env
+       - ../another/dev.env
+```
+
+Local project environment have precendence over values set in this file, so that local project can
+override values for customization.
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -17,6 +17,15 @@
       "description": "define the Compose project name, until user defines one explicitly."
     },
 
+    "include": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "$ref": "#/definitions/include"
+      },
+      "description": "compose sub-projects to be included."
+    },
+
     "services": {
       "id": "#/properties/services",
       "type": "object",
@@ -586,6 +595,22 @@
         "additionalProperties": false,
         "patternProperties": {"^x-": {}}
       }
+    },
+
+    "include": {
+      "id": "#/definitions/include",
+      "oneOf": [
+        {"type": "string"},
+        {
+          "type": "object",
+          "properties": {
+            "path": {"$ref": "#/definitions/string_or_list"},
+            "env_file": {"$ref": "#/definitions/string_or_list"},
+            "project_directory": {"type": "string"}
+          },
+          "additionalProperties": false
+        }
+      ]
     },
 
     "network": {
--- a/spec.md
+++ b/spec.md
@@ -2882,3 +2882,93 @@ services:
   foo:
     build: null
 ```
+## Dependency on other compose projects
+
+A Compose application can declare dependency on another compose application, managed by another team
+or shared with others. This allows to keep a compose file reasonably complicated for the limited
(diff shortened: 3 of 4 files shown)


spec.md
```diff
@@ -2888,3 +2888,32 @@ services:
...
     build: null
 ```
+## Dependency on other compose projects
+
+A Compose project can declare dependency on another compose project, managed by another team
+or shared with others. This allows to keep a compse file reasonably complicated for the limited
+amount of resources a single team has to declare for it's own sub-domain within a larger deployment.
+
+`require` top-level section is used to define dependency to another compose application (or subdomain).
+Each path listed in this section MUST be loaded as an individual compose model with it's own project
```
nicks, line 2898: are you also imagining that each individual compose model will have it's own .env file and do it's own env resolution?

spec.md
```diff
@@ -2888,3 +2888,32 @@ services:
...
+
+```yaml
+require:
+  - ../commons/compose.yaml
+  - ../another_domain/compose.yaml
+
+services:
+  webapp:
+    depends_on:
+      - imported-service # defined by another_domain
```
nicks, line 2910: i'm a little bit worried about automatically importing all of "another_domain"s names into the main project's name space. When I was prototyping this with real projects, I ran into this problem very quickly:

User Service Compose File:
```
services:
  api: ...
  db: ...
```

Repo Service Compose File
```
services:
  api: ...
  db: ...
requires:
- ./users.compose.yaml
```

where you really don't want to merge the user DB and repo DB together, you want them to be separate DBs. 

I feel like we're going to need something more like:

```
imports:
  - from: ./users.compose.yaml
    service: api 
    as: users-api
```

where both the users project and repos project both get loaded, but they have their own networks and we have ways to link them without overwriting names.

spec.md
```diff
@@ -2888,3 +2888,32 @@ services:
   foo:
     build: null
 ```
+## Dependency on other compose projects
+
+A Compose project can declare dependency on another compose project, managed by another team
+or shared with others. This allows to keep a compse file reasonably complicated for the limited
```
nicks, line 2894:
```suggestion
or shared with others. This allows to keep a compose file reasonably complicated for the limited
```

spec.md
ndeloof: if you only need a single service, especially if you need to rename or customize, better use `extends` (which was the conclusion on https://github.com/docker/compose/issues/318)

spec.md
ndeloof: good point, didn't considered this (which is out of the compose spec scope)

I tried to make this minimalist on first run, but we obviously could have a "long syntax" to offer fine grain control on resources we import into current model:
```yaml
include:
   - source: ../another/compose.yaml
     project-directory: /tmp
     excludes: ["db"]
     rename: foo=bar
     env: ../another/.env
```

spec.md
ndeloof: `extends` is designed for this use case:

```yaml
services:
  users-api:
    extends: 
      file : ./users.compose.yaml
      service: api 
```

nicks: hmm...just tried that and it didn't behave how i'm describing, so i think there's a misunderstanding.

in my use case, i want to pull in the whole Users project.

users.compose.yaml is maintained by another team. I don't know what services they need to run. They should be able to add services to their compose file without breaking the main project.

In your example, extends only pulls in the api service, not the other services in the file. changes to the users.compose.yaml file would break all the extending projects.

spec.md
nicks: :+1: ya, i appreciate the update. just make it an error if the names clash for now, and we can iterate later.

spec.md
ndeloof: right, then this is the exact goal for this feature. And corner cases like includes/excludes or renaming can be implemented as a second interation

ndeloof:
long syntax indeed let us introduce way much advanced features in the future to address complex usages, like includes/excludes/rename or any sort of such conflict resolution logic.

ulyssessouza (changes requested):
I think one important thing to mention here is the order in what `-f` files and `compose.override.yml` are merged (which already exists) and when the imports are evaluated.

Also maybe turn it explicit the recursive nature of the `import`. Because if we import a file that has imports, we need to evaluate those before integrating in the final project. And obviously manage circular dependency.

ndeloof:
@ulyssessouza good point, added to https://github.com/compose-spec/compose-spec/pull/363
(those comment should apply to the spec discussion)

About circular references, the compose specification doesn't define such rule for `extends` not `depends_on` while those are actually implemented in Docker Compose. We could look into this in a separate PR

14-include.md
```diff
@@ -0,0 +1,89 @@
+## Dependency on other compose projects
+
+A Compose project can declare dependency on another compose project, managed by another team
```
glours, line 3: I didn't find any reference to `Compose project` in the specification, to be consistent we should use `application` or `configuration`
```suggestion
A Compose application can declare dependency on another compose application, managed by another team
```

14-include.md
```diff
@@ -0,0 +1,89 @@
...
+or shared with others. This allows to keep a compose file reasonably complicated for the limited
+amount of resources a single team has to declare for it's own sub-domain within a larger deployment.
+
+`include` top-level section is used to define dependency to another compose application (or subdomain).
+Each path listed in this section MUST be loaded as an individual compose model with it's own project
```
glours, line 8: Regarding this [PR](https://github.com/compose-spec/compose-spec/pull/370), we should already apply the same format.
```suggestion
Each path listed in this section must be loaded as an individual compose model with it's own project
```

14-include.md
```diff
@@ -0,0 +1,89 @@
...
+directory to resolve relative paths. 
+
+Once project to be included has been loaded, all resources definitions are copied into the
```
glours, line 11:
```suggestion
Once the project to be included has been loaded, all resources definitions are copied into the 
```

14-include.md
```diff
@@ -0,0 +1,89 @@
...
+A Compose project can declare dependency on another compose project, managed by another team
+or shared with others. This allows to keep a compose file reasonably complicated for the limited
+amount of resources a single team has to declare for it's own sub-domain within a larger deployment.
+
+`include` top-level section is used to define dependency to another compose application (or subdomain).
+Each path listed in this section MUST be loaded as an individual compose model with it's own project
+directory to resolve relative paths. 
+
+Once project to be included has been loaded, all resources definitions are copied into the 
+current compose application model. A Compose implementation MUST warn user if some resource
```
glours, line 12:
```suggestion
current compose application model. Compose must warn user if some resource 
```

14-include.md
```diff
@@ -0,0 +1,89 @@
...
+or shared with others. This allows to keep a compose file reasonably complicated for the limited
+amount of resources a single team has to declare for it's own sub-domain within a larger deployment.
+
+`include` top-level section is used to define dependency to another compose application (or subdomain).
+Each path listed in this section MUST be loaded as an individual compose model with it's own project
+directory to resolve relative paths. 
+
+Once project to be included has been loaded, all resources definitions are copied into the 
+current compose application model. A Compose implementation MUST warn user if some resource 
+name conflict, and not try to merge those. To enforce this, `include` MUST be evaluated after the compose file(s) selected by user to define the Compose application model have been parsed and merged,
```
glours, line 13:
```suggestion
name conflict, and not try to merge those. To enforce this, `include` must be evaluated after the compose file(s) selected by user to define the Compose application model have been parsed and merged,
```

14-include.md
```diff
@@ -0,0 +1,89 @@
...
+
+```yaml
+include:
+   - path: ../commons/compose.yaml
+     project_directory: ..
+     env_file: ../another/.env
+```
+
+#### path
+`path` is REQUIRED and defines the location of the compose file(s) to be parsed and included into
```
glours, line 56:
```suggestion
`path` is required and defines the location of the compose file(s) to be parsed and included into
```

glours (approved):
If you can apply those few changes I think we're ready to go!!!

schema/compose-spec.json
```diff
@@ -588,6 +588,22 @@
       }
     },
 
+    "require": {
+      "id": "#/definitions/require",
```
milas, line 592:
```suggestion
    "include": {
      "id": "#/definitions/include",
```

schema/compose-spec.json
```diff
@@ -588,6 +588,22 @@
...
+      "oneOf": [
+        {"type": "string"},
+        {
+          "type": "object",
```
milas, line 596: I think this needs to be `"type": "array"` and then this definition would be for `"items"`

milas (approved):
Nice clarifications to the spec. I think the JSON schema needs a couple tiny fixes but LGTM otherwise!

schema/compose-spec.json
ndeloof: yes indeed, this is incomplete copy/paste from https://github.com/compose-spec/compose-go/pull/416/files#diff-5f5b95e68cadf606dc6b4956ca67b87039e3283cb01aa3208fda399abc167906

ulyssessouza (approved):

reddec:
May I ask for clarification? 
In the case of a `default` network (ie, not defined explicitly), all services will share the network space unless changed by the `- networks` parameter, right?

Example:

`foo/docker-compose.yaml`:

```yaml
services:
  foo:
    image: busybox
```


`bar/docker-compose.yaml`:


```yaml
services:
  bar:
    image: busybox
```


`docker-compose.yaml`:

```yaml
include:
- foo/docker-compose.yaml
- bar/docker-compose.yaml
```


`bar` and `foo` services will share the same `default` network and should be to communicate with each other (including hostname links).

PR #364, opened 2023-06-12
introduce remove to configure runtime autoremove for service containers

ndeloof:
introduce `remove` to configure runtime autoremove for service containers

Fixes https://github.com/docker/compose/issues/10688

### DIFF
--- a/05-services.md
+++ b/05-services.md
@@ -1393,6 +1393,14 @@ If `pull_policy` and `build` are both present, Compose implementations SHOULD bu
 
 `read_only` configures service container to be created with a read-only filesystem.
 
+### remove
+
+When set to `true`, `remove` will configure platform to automatically remove the
+resource when the process exits. This has no effect if `restart` is set.
+
+Service continuity MAY be affected, as resource is removed before Compose implementation
+can react to process termination. Typically, anonymous volumes will be lost.
+
 ### restart
 
 `restart` defines the policy that the platform will apply on container termination.
--- a/spec.md
+++ b/spec.md
@@ -1662,6 +1662,14 @@ If `pull_policy` and `build` are both present, Compose implementations SHOULD bu
 
 `read_only` configures service container to be created with a read-only filesystem.
 
+### remove
+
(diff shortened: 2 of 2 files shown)


05-services.md
```diff
@@ -1393,6 +1393,14 @@ If `pull_policy` and `build` are both present, Compose implementations SHOULD bu
 
 `read_only` configures service container to be created with a read-only filesystem.
 
+### remove
+
+When set to `true`, `remove` will configure platform to automatically remove the
+resource when the process exits. This has no effect if `restart` is set.
+
+Service continuity MAY be affected, as resource is removed before Compose implementation
+can react to process termination. Typically, anonymous volumes will be lost.
```
milas, line 1402: > before Compose implementation can react to process termination

Should we disallow having `depends_on` reference a service that has `remove: true`? Or is there something else you're concerned about here?

(Sidenote: is it possible to define anonymous volumes in Compose?)

milas (approved):
I'm slightly nervous about folks using this without understanding and then wondering what happened to their services 😅, but at the same time, it's a really handy feature in the Engine, so I support exposing it through Compose for parity.

(And I can see it'd be useful for sidecar purposes or tasks/jobs.)

05-services.md
ndeloof: The challenge is not about dependencies but about container restart & replacement. Inheriting anonymous volumes is the first to come in my mind but there's probably a few other corner-cases.
I can see benefit of this feature with container used with `docker compose run` to avoid orphaned. For services used with `up` this is not so obvious.

> Sidenote: is it possible to define anonymous volumes in Compose

nope, anonymous are only defined by image metadata (Dockerfile's `VOLUME`)  when no explicit binding is defined in container config.

jhrotko:
@ndeloof I assume this PR does not make sense anymore as the source of truth is compose-go?

ndeloof:
forgot about this one. Let's close this as obsolete I'll revisit later

PR #366, opened 2023-06-13
introduce `service.attach`

ndeloof:
introduce `attach` so a verbose utility service can be configured to be ignored when collecting output by setting `attach: false`. Defaults to `attach: true`.

demonstration usage on https://github.com/docker/compose/pull/10700

see https://github.com/docker/compose/issues/10695

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -235,6 +235,13 @@ Deploy support is an optional aspect of the Compose specification, and is
 described in detail in the [Deployment support](deploy.md) documentation.
 If not implemented the Deploy section should be ignored and the Compose file must still be considered valid.
 
+### attach
+
+When `attach` is defined and set to `false` Compose will not collect service logs,
+until user explicitly requests to.
+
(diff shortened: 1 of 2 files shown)


glours (approved):
LGTM 👍

spec.md
```diff
@@ -294,6 +294,13 @@ Deploy support is an OPTIONAL aspect of the Compose specification, and is
 described in detail in the [Deployment support](deploy.md) documentation.
 If not implemented the Deploy section SHOULD be ignored and the Compose file MUST still be considered valid.
 
+### attach
+
+When `attach` is defined and set to `false` a Compose implementation SHOULD not collect service logs,
```
laurazard, line 299: We should align this wording with the new tone, and remove `Compose implementation` and the uppercase `SHOULD`

laurazard (approved):
LGTM, just a comment

PR #372, opened 2023-06-23
ENGDOCS-1305

aevesdocker:
This PR:
- aligns `14-include.md` with style guide and clarifies the meaning of some sentences.
- splits the `03-compose.md` file into two pages and moves the profiles section into it's own .md file. This allows for greater linkages off of the compose file page to useful pages such as fragments, extensions, merge and include, and elevates the profile content as it was a bit buried.
- aligns `02-model.md`, `03-compose-file.md` and 15-profiles.md` with the style guide

### DIFF
--- a/02-model.md
+++ b/02-model.md
@@ -12,7 +12,9 @@ Some services require configuration data that is dependent on the runtime or pla
 
 A [secret](09-secrets.md) is a specific flavor of configuration data for sensitive data that should not be exposed without security considerations. Secrets are made available to services as files mounted into their containers, but the platform-specific resources to provide sensitive data are specific enough to deserve a distinct concept and definition within the Compose specification.
 
-Distinction within Volumes, Configs and Secret allows implementations to offer a comparable abstraction at service level, but cover the specific configuration of adequate platform resources for well identified data usages.
+>**Note**
+>
+> With volumes, configs and secrets you can have a simple declaration at the top-level and then add more platform-specific information at the service level.
 
 A project is an individual deployment of an application specification on a platform. A project's name, set with the the top-level [`name`](04-verision-and-name.md) attribute, is used to group
 resources together and isolate them from other applications or other installation of the same Compose specified application with distinct parameters. If you are creating resources on a platform, you must prefix resource names by project and
--- a/02-model.md
+++ b/02-model.md
@@ -24,7 +24,7 @@ Project names must contain only lowercase letters, decimal digits, dashes, and u
 
 ### Illustrative example
 
-The following example illustrates the Compose Specification concepts outlined. The example is non-normative.
+The following example illustrates the Compose Specification concepts outlined above. The example is non-normative.
 
 Consider an application split into a frontend web application and a backend service.
 
--- a/02-model.md
+++ b/02-model.md
@@ -1,30 +1,30 @@
 ## The Compose application model
 
-The Compose specification allows one to define a platform-agnostic container based application. Such an application is designed as a set of containers which have to both run together with adequate shared resources and communication channels.
+The Compose Specification lets you define a platform-agnostic container based application. Such an application is designed as a set of containers which have to both run together with adequate shared resources and communication channels.
 
-Computing components of an application are defined as [Services](05-services.md). A Service is an abstract concept implemented on platforms by running the same container image (and configuration) one or more times.
+Computing components of an application are defined as [services](05-services.md). A service is an abstract concept implemented on platforms by running the same container image, and configuration, one or more times.
 
-Services communicate with each other through [Networks](06-networks.md). In this specification, a Network is a platform capability abstraction to establish an IP route between containers within services connected together. Low-level, platform-specific networking options are grouped into the Network definition and may be partially implemented on some platforms.
+Services communicate with each other through [networks](06-networks.md). In the Compose Specification, a network is a platform capability abstraction to establish an IP route between containers within services connected together. Low-level, platform-specific networking options are grouped into the Network definition and may be partially implemented on some platforms.
 
-Services store and share persistent data into [Volumes](07-volumes.md). The specification describes such a persistent data as a high-level filesystem mount with global options. Actual platform-specific implementation details are grouped into the Volumes definition and may be partially implemented on some platforms.
+Services store and share persistent data into [volumes](07-volumes.md). The Specification describes such a persistent data as a high-level filesystem mount with global options. Actual platform-specific implementation details are grouped into the volumes definition and may be partially implemented on some platforms.
 
-Some services require configuration data that is dependent on the runtime or platform. For this, the specification defines a dedicated concept - [Configs](08-configs.md). From a Service container point of view, Configs are comparable to Volumes, in that they are files mounted into the container. But the actual definition involves distinct platform resources and services, which are abstracted by this type.
+Some services require configuration data that is dependent on the runtime or platform. For this, the Specification defines a dedicated [configs](08-configs.md) concept. From a service container point of view, configs are comparable to volumes, in that they are files mounted into the container. But the actual definition involves distinct platform resources and services, which are abstracted by this type.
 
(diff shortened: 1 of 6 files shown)


14-include.md
```diff
@@ -1,29 +1,27 @@
...
+current Compose application model. Compose displays a warning if resource names conflict and doesn't 
+try to merge them. To enforce this, `include` is evaluated after the Compose file(s) selected 
+to define the Compose application model have been parsed and merged, so that conflicts 
+between Compose files are detected.
 
-`include` applies recursively: an included compose file which declares it's own `include` section
-will trigger those other files to be included as well.
+`include` applies recursively so an included Compose file which declares its own `include` section, triggers those other files to be included as well.
 
 The resulting resources can be used in the including compose model for cross-services references.
```
aevesdocker, line 18: TODO: Check the exact meaning of this sentence and reword

02-model.md
```diff
@@ -1,38 +1,38 @@
...
 
-A [Secret](09-secrets.md) is a specific flavor of configuration data for sensitive data that should not be exposed without security considerations. Secrets are made available to services as files mounted into their containers, but the platform-specific resources to provide sensitive data are specific enough to deserve a distinct concept and definition within the Compose specification.
+A [secret](09-secrets.md) is a specific flavor of configuration data for sensitive data that should not be exposed without security considerations. Secrets are made available to services as files mounted into their containers, but the platform-specific resources to provide sensitive data are specific enough to deserve a distinct concept and definition within the Compose specification.
 
 Distinction within Volumes, Configs and Secret allows implementations to offer a comparable abstraction at service level, but cover the specific configuration of adequate platform resources for well identified data usages.
```
aevesdocker, line 15: TODO: Check the meaning of this sentence and reword

15-profiles.md
```diff
@@ -0,0 +1,61 @@
...
+  zot:
+    image: zot
+    depends_on:
+      - bar
+    profiles:
+      - debug
+```
+
+In the above example: 
+- If the Compose application model is parsed with no profile enabled, it  only contains the `foo` service.
```
craig-osterhout, line 45:
```suggestion
- If the Compose application model is parsed with no profile enabled, it only contains the `foo` service.
```

14-include.md
```diff
@@ -1,29 +1,27 @@
...
 
-### short syntax
+### Short syntax
 
-Short syntax only defines path to another compose file. File is loaded with parent
-folder as project directory, and optional `.env` file being loaded to define variables default values
-for interpolation, while local project environment can override those values. 
+The short syntax only defines paths to another Compose file. The file is loaded with the parent
```
craig-osterhout, line 22:
```suggestion
The short syntax only defines paths to other Compose files. Each file is loaded with the parent
```
It currently reads like multiple paths to a single file because plural `paths` but singular `Compose file`. Not sure if the suggested change is correct.

02-model.md
glours: Unlike `Configuration`, `Secrets` can be pass also via environment variables, they're also mounted into a specific directory in the container `/run/secrets`... This 2 main differences explains why `Secrets` needs its own type

14-include.md
aevesdocker: Ah I wasn't clear - I mean for line 18, not 16.

02-model.md
aevesdocker: Also i need an explanation for line 15 note 13! GitHub isn't good at making clear which line I'm referring to!

14-include.md
glours: Ok sorry, so it means that the services, volumes, networks ... which are declared in the Compose files included can be used as regular resources in the current Compose file
for example
```
include:
  - my-compose-include.yaml  #with serviceB declared for example
services:
  serviceA:
    build: .
    depends_on:
      - serviceB #user serviceB directly like it was declared in this Compose file
 ```

File: 02-model.md (hunk 2 above)
Comment (glours) [reply 02-model.md:15]: On a the service level, you can declare a `volume`, `secrets` or `configs` or detail existing ones with info specific to the platform, example from [secret declaration on service level](https://github.com/compose-spec/compose-spec/blob/master/05-services.md#long-syntax-4): 
```
services:
  frontend:
    image: awesome/webapp
    secrets:
      - source: server-certificate
        target: server.cert
        uid: "103"
        gid: "103"
        mode: 0440
secrets:
  server-certificate:
    external: true
```

As you can see, you could define a simple declaration on top secret element and add more details in the service.

File: 14-include.md (hunk 4 above)
Comment (aevesdocker) [reply 14-include.md:32]: ah yes! good catch

File: 14-include.md (hunk 1 above)
Comment (aevesdocker) [reply 14-include.md]: Got it! Thanks for clarifying. Have changed the wording and actually added your example above coz I think it is needed

File: 02-model.md (hunk 2 above)
Comment (aevesdocker) [reply 02-model.md]: Comprende. I've changed the wording but may have simplified it too much now. LMK

**Comment by aevesdocker** [comment]
cc @ndeloof

File: 03-compose-file.md [hunk 5]
```diff
@@ -1,78 +1,23 @@
...
-- If Compose is executed with `baz` as explicit service to run, the service `baz` and the
-  profile `test` will be active and `bar` will be pulled in by the `depends_on` constraint.
-- If Compose is executed with `zot` as explicit service to run, again the model will be
-  invalid regarding the `depends_on` constraint of `zot`, since `zot` and `bar` have no common `profiles`
-  listed.
-- If Compose is executed with `zot` as explicit service to run and profile `test` enabled,
-  profile `debug` is automatically enabled and service `bar` is pulled in as a dependency starting both
-  services `zot` and `bar`.
-
+You can also use [`include`](14-include.md) in your Compose file if you want your Compose application to be dependent on another Compose application that is managed by a different team or shared with others.
```
laurazard, line 23: Same here:
```suggestion
If you want to reuse other Compose files or factor out parts of your application model into separate Compose files, you can also use [`include`](14-include.md).
This is useful in situations where your Compose application is dependent on another application which is managed by a different team, or needs to be shared with others.
```
I think this reads better, and explains what `include` does and what it's useful for.

laurazard (commented):
LGTM, left a few nits re: the new wording for the short paragraph on `include` and on the overall `include` description in `14-include.md`

03-compose-file.md
aevesdocker: Agreed! Changed 👍

14-include.md
```diff
@@ -1,29 +1,37 @@
-## Dependency on other compose projects
+## Include
 
-A Compose application can declare dependency on another compose application, managed by another team
-or shared with others. This allows to keep a compose file reasonably complicated for the limited
-amount of resources a single team has to declare for it's own sub-domain within a larger deployment.
+A Compose application can declare dependency on another Compose application that is managed by another team
+or shared with others. This is useful if teams need to keep a Compose file reasonably complicated for the limited
+amount of resources it has to declare for it's own sub-domain, within a larger deployment.
```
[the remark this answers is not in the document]
aevesdocker: Nice! I've added this. I changed the formate a bit so less long sentences, more easy-to-read bullet points

aevesdocker:
Thank you @laurazard! Made the changes, ready for re-review

laurazard (approved):

14-include.md
```diff
@@ -1,29 +1,39 @@
...
+## Include
 
-A Compose application can declare dependency on another compose application, managed by another team
-or shared with others. This allows to keep a compose file reasonably complicated for the limited
-amount of resources a single team has to declare for it's own sub-domain within a larger deployment.
+A Compose application can declare dependency on another Compose application that is managed by another team
+or shared with others. This is useful if:
+- You want to reuse other Compose files
+- You need to factor out parts of your application model into separate Compose files so they can be managed separately or shared with others.
+- Teams need to keep a Compose file reasonably complicated for the limited amount of resources it has to declare for it's own sub-domain, within a larger deployment.
```
laurazard, line 7: Thanks for looking at the suggestions @aevesdocker! I still think here, we should have a more clear delineation – A Compose file can declare dependency on another Compose file [full stop]. Whether that other Compose application is managed by another team or shared with others is irrelevant to whether a Compose application [**can**] declare dependency, and should live in the bullet points where it's explained where this can be useful.

glours (approved):

PR #376, opened 2023-06-28
make `build.context` optional and default to `.`

ndeloof:
make `build.context` optional and default to `.`

### DIFF
--- a/build.md
+++ b/build.md
@@ -100,7 +100,7 @@ The second part represents a subdirectory inside the repository that is used as 
 
 Alternatively `build` can be an object with fields defined as follows:
 
-### context (Required)
+### context
 
 `context` defines either a path to a directory containing a Dockerfile, or a URL to a git repository.
 
@@ -118,6 +118,8 @@ services:
   webapp:
     build: https://github.com/mycompany/webapp.git
 ```
+
+If not set explicitly, `context` defaults to project directory (`.`). 
 
 ### dockerfile
 


glours (approved):
LGTM

milas (approved):
LGTM!

This has been the behavior (unintentionally) of Compose v2.x all along, and nobody has complained about it...until we "fixed" it in v2.19, so I think this is a reasonable relaxing of strictness.

See also:
 * compose-spec/compose-go#423
 * compose-spec/compose-go#424

danfundnel:
@ndeloof I may be wrong but it seems this change may have had some side-effects (not sure if intended) in the case where `extends:` is used to import compose specs from files in other dirs which specify a relative `build.context` path. Up until the 4.21 upgrade compose would take the reference in the included compose spec as relative to that file. It now seems to require it to be set relative to the directory docker-compose is called in (and, by the way, not the dir of the parent compose file using `extends:`, if that file is specified using `-f`). I think either behaviour is fine but this has broken the specs in our monorepo so I imagine others may also have to make this change.

ndeloof:
yes indeed, see https://github.com/docker/compose/issues/10771

ndeloof:
@danfundnel I don't think so
quick experiment:
compose.yaml
```yaml
services:
  app:
    extends:
      file: ./abstract/base.yaml
      service: base
```
base.yaml
```yaml
services:
  base:
    build:
      context: ./base
```

```
$ docker compose config
services:
  app:
    build:
      context: /Users/nicolas/test/abstract/base
      dockerfile: Dockerfile
```

PR #382, opened 2023-07-04
add required attribute depends_on

### LINKED ISSUE #274: Allow depends_on to be optional to use profiles to control which services to run in development vs production

We can't use profiles to enable specific services in development but not production when using `depends_on`. That's because the current spec doesn't allow `depends_on` to be optional. This means in practice profiles can't be used to enable specific services in different environments which is the goal of profiles based on its first sentence of the official documentation.

In practice this is a very big deal because in development you may want to use `postgres` and `redis` locally through Docker Compose but in production you may use managed cloud provider services for this. You could say our apps do have dependencies on these services but they are optional from Docker Compose's point of view.

I've already gone into a huge amount of detail about this in https://github.com/docker/compose/issues/9795 but here's a recap:

Imagine this set up. This actually works with v2.10.2 because there's supposedly an unfixed bug but it turns out this bug is actually what folks have been wanting for years since profiles were available.

```yml
services:
  postgres:
    profiles: ["postgres"]
  redis:
    profiles: ["redis"]
  web:
    profiles: ["web"]
    depends_on: ["postgres", "redis"]
  worker:
    profiles: ["worker"]
    depends_on: ["postgres", "redis"]
  assets:
    profiles: ["assets"]
```

In development you can set this environment variable in an ` .env` file to start everything:

```export COMPOSE_PROFILES=postgres,redis,assets,web,worker```

In production you can set this instead:

```export COMPOSE_PROFILES=web,worker```

This would let you avoid running `assets` containers in production for various watchers and also take advantage of using your cloud provider's managed Postgres and Redis services because the local containers for those services don't need to run since the cloud provider's managed service will fulfill that dependency.

The above is a fantastic solution to let developers pick and choose which services to run. The problem is the current spec doesn't allow the above. Docker Compose will error out saying your dependent services aren't available.

One person opened an issue bringing that up and they want it fixed to be more inline with the spec stating it's confusing. Many alternative solutions were proposed in https://github.com/docker/compose/issues/9795, some of which were to use multiple docker-compose.yml files or introducing more complexity into Docker Compose with competing options like `COMPOSE_SERVICE_FILTER` to allow for services to be defined and always running commands with `--no-deps`. None of which are ideal or allow the desired behavior.

**A proposed solution**

We already have profiles which according to the docs states:

> Profiles allow adjusting the Compose application model for various usages and environments by selectively enabling services

What about adding a new `necessity: "optional"` property which defaults to `necessity: "required"` to `depends_on` so it can all work the same with profiles being explicit, or maybe `required: false` which defaults to `required: true`?  That's in my opinion the easiest solution and it's backwards compatible for the "fixed" behavior while giving people who have already depended on the "unfixed" behavior a way to use Docker Compose in the way v2 has allowed us to use since the beginning of the year.

I'd really like this to be implemented before https://github.com/docker/compose/issues/9795 is addressed. Otherwise Docker Compose v2 will introduce breaking changes with this fix since this unfixed behavior has been available since v2 was around which has been almost a year? It was even GA'd almost 6 months ago.

There is currently no workaround other than using `sed` and other tools to programmatically alter a `docker-compose.yml` file for each project before the project is upped. This has been a workaround I've used for years because it was the only way to get the desired behavior of disabling `depends_on` on a few services in production while still using a single `docker-compose.yml` file for all environments. It's hacky, cryptic, error prone and required for every project.

The proposed solution here fixes everything to where you can enable specific services in any environment while still being able to use `depends_on` optionally. It sticks to the goal of what profiles says it does based on its documentation and improves the quality of life for developers and overall improves the developer experience.

### PR DESCRIPTION
**Pull request by glours**
Introduce a `required` attribut to `depends_on` to allow a Compose implementation to only warn user when a dependency service isn't started or available.

Why this change? 
Since the beginning the reference implementation Docker Compose was silently ignore missing dependency services, recently a fix was introduced to strictly follow the specification and broke existing users Compose configurations. 
Adding this new attribute will let them continue to use the previous behaviour while making clear and explicit the `depends_on` could be bypass by the Compose implementation

Fixes #274

### DIFF
--- a/05-services.md
+++ b/05-services.md
@@ -386,6 +386,8 @@ expressed in the short form.
     service.
   - `service_completed_successfully`: specifies that a dependency is expected to run
     to successful completion before starting a dependent service.
+- `required`: when `false` Compose only warns you when the dependency service isn't started or available. If not defined
+    the default value of `required` is `true`.
 
 Service dependencies cause the following behaviors:
 
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -191,6 +191,10 @@
                   "additionalProperties": false,
                   "properties": {
                     "restart": {"type": "boolean"},
+                    "required": {
+                      "type":  "boolean",
+                      "default": true
+                    },
                     "condition": {
                       "type": "string",
                       "enum": ["service_started", "service_healthy", "service_completed_successfully"]
--- a/spec.md
+++ b/spec.md
@@ -597,6 +597,8 @@ expressed in the short form.
     service.
   - `service_completed_successfully`: specifies that a dependency is expected to run
     to successful completion before starting a dependent service.
+- `required`: when `false` Compose only warns you when the dependency service isn't started or available. If not defined
+    the default value of `required` is `true`.
 
 Service dependencies cause the following behaviors:
 

### CODE REVIEW

File: 05-services.md [hunk 1]
```diff
@@ -379,6 +379,8 @@ expressed in the short form.
     service.
   - `service_completed_successfully`: specifies that a dependency is expected to run
     to successful completion before starting a dependent service.
+- `required`: when `false` Compose only warns you when the dependency service isn't started or available. If not defined
+    the default value of `required` will be `true`.
```
ndeloof, line 383:
```suggestion
    the default value of `required` is `true`.
```

ndeloof (approved):

05-services.md
```diff
@@ -379,6 +379,8 @@ expressed in the short form.
     service.
   - `service_completed_successfully`: specifies that a dependency is expected to run
     to successful completion before starting a dependent service.
+- `required`: when `false` Compose only warns you when the dependency service isn't started or available. If not defined
```
laurazard, line 382: Nit: maybe an empty line such as in 374 to separate the new field description more from the rest?

laurazard (approved):
LGTM! Just left a nit.

PR #383, opened 2023-07-06
Add `start_interval` to healthcheck

cpuguy83:

Start interval is the interval that healthchecks are performed during the start period.
This was added in moby in https://github.com/moby/moby/pull/40894.
You can see the linked issue in that PR for the number of users asking for this.

Docker CLI changes: https://github.com/docker/cli/pull/4405
Swarmkit changes: https://github.com/moby/swarmkit/pull/3142

### DIFF
--- a/05-services.md
+++ b/05-services.md
@@ -896,9 +896,10 @@ healthcheck:
   timeout: 10s
   retries: 3
   start_period: 40s
+  start_interval: 5s
 ```
 
-`interval`, `timeout` and `start_period` are [specified as durations](11-extension.md#specifying-durations).
+`interval`, `timeout`, `start_period`, and `start_interval` are [specified as durations](11-extension.md#specifying-durations).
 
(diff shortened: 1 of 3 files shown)


glours (approved):

laurazard:
It's finally here 🎉 ! I'm just looking at this from my phone, but I wonder if we need changes to the schema json (somewhere around https://github.com/compose-spec/compose-spec/blob/ec66400775028485e4d3fbcf8d3f0b84578edd52/schema/compose-spec.json#L460)?

glours:
You're right @laurazard 👍 , we need to update `compose-spec.json` schema file

laurazard:
@cpuguy83 can you take care of that or should one of us make some quick changes?

cpuguy83:
Updated

laurazard (approved):
Thanks ❤️

PR #385, opened 2023-07-12
volumes.md tidy

aevesdocker:
This PR aligns volumes.md with the style guide. It also provides greater clarity on how to use the volumes top-level element and improves the flow in some places.

### DIFF
--- a/07-volumes.md
+++ b/07-volumes.md
@@ -25,9 +25,9 @@ volumes:
   db-data:
 ```
 
-The `db-data` volume is stored at `/var/lib/backup/data` and `/etc/data` container paths for backup and backend respectively.
+The `db-data` volume is mounted at the `/var/lib/backup/data` and `/etc/data` container paths for backup and backend respectively.
 
-When `docker compose up` is run for the firs time, it creates the volume. Docker reuses the same volume when you run the command subsequently.
+Running `docker compose up` creates the volume if it doesn't already exist. Otherwise, the existing volume is used and is recreated if it's manually deleted outside of Compose.
 
 ### Attributes
 
@@ -84,7 +84,7 @@ volumes:
 
 `labels` are used to add metadata to volumes. You can use either an array or a dictionary.
 
-Use reverse-DNS notation to prevent your labels from conflicting with those used by other software.
+It's recommended that you use reverse-DNS notation to prevent your labels from conflicting with those used by other software.
 
 ```yml
 volumes:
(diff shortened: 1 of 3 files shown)


ndeloof (approved):

07-volumes.md
```diff
@@ -23,22 +25,28 @@ volumes:
   db-data:
 ```
 
-An entry under the top-level `volumes` key can be empty, in which case it uses the platform's default configuration for
+The `db-data` volume is stored at `/var/lib/backup/data` and `/etc/data` container paths for backup and backend respectively.
```
laurazard, line 28:
```suggestion
The `db-data` volume is mounted at the `/var/lib/backup/data` and `/etc/data` container paths for backup and backend respectively.
```

07-volumes.md
```diff
@@ -23,22 +25,28 @@ volumes:
...
+
+When `docker compose up` is run for the firs time, it creates the volume. Docker reuses the same volume when you run the command subsequently.
```
laurazard, line 30: typo:
```suggestion
When `docker compose up` is run for the first time, it creates the volume. Docker reuses the same volume when you run the command subsequently.
```

But also I think this paragraph could be more explicit:
```suggestion
`docker compose up` will create the volume if it doesn't already exist (such as on first execution). Otherwise, the existing volume is used (and will be recreated if it is manually deleted outside of Compose).
```

07-volumes.md
```diff
@@ -75,12 +80,11 @@ volumes:
...
 ```
 
-### labels
+#### labels
 
 `labels` are used to add metadata to volumes. You can use either an array or a dictionary.
 
-It's recommended that you use reverse-DNS notation to prevent your labels from
-conflicting with those used by other software.
+Use reverse-DNS notation to prevent your labels from conflicting with those used by other software.
```
laurazard, line 87: I don't think omitting the "it's recommended" part is good, here. Any other input here is valid, and that could be confusing if we don't call that out. This sentence is explicitly a usage recommendation, not a prescriptive definition.

laurazard (changes requested):
Some typos/nits/suggestions

aevesdocker:
Thanks @laurazard! I've incorporated your comments

laurazard:
Thanks @aevesdocker, LGTM! Can you squash those commits and we can get it merged?

aevesdocker:
@laurazard I have the fancy 'squash and merge' button here, so can do that once you've approved ✅ 😃 

Screenshot 2023-07-19 at 13 31 17

laurazard (approved):

PR #399, opened 2023-08-07
ENGDOCS-1560

aevesdocker:
Adds requested info from the Community Slack


glours (approved):
👍

aevesdocker:
@milas Would be good to perhaps add some info as to why this would useful? Why would someone want to do this?

laurazard (approved):
LGTM

PR #404, opened 2023-08-17
introduce depends_on.service to define the target service

ndeloof:
introduce `depends_on.service` to define target service. Defaults to mapping key, but could be overridden, typically using a variable.

see https://github.com/docker/compose/issues/10855

### DIFF
--- a/05-services.md
+++ b/05-services.md
@@ -380,6 +380,7 @@ expressed in the short form.
     to successful completion before starting a dependent service.
 - `required`: When set to `false` Compose only warns you when the dependency service isn't started or available. If it's not defined
     the default value of `required` is `true`.
+- `service`: name of the target service (defaults to mapping key). Value can be set by a variable.
 
 Service dependencies cause the following behaviors:
 
@@ -398,7 +399,8 @@ services:
   web:
     build: .
     depends_on:
-      db:
+      database:
+        service: db
         condition: service_healthy
         restart: true
       redis:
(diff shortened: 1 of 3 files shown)


laurazard:
Is this still relevant? If so, can we maybe think about other names for this property other than `service`? Maybe `target`?

ndeloof:
I don't feel `depends_on.target: XX` to be more readable than `depends_on.service: XX`

05-services.md
```diff
@@ -398,7 +399,8 @@ services:
   web:
     build: .
     depends_on:
-      db:
+      database:
+        service: db
```
laurazard, line 403: I think only having this example might be a bit confusing for new users who don't know about this feature. I'd leave this example as is, and add a specific example about the `service` key under it's explanation

05-services.md
```diff
@@ -380,6 +380,7 @@ expressed in the short form.
     to successful completion before starting a dependent service.
 - `required`: When set to `false` Compose only warns you when the dependency service isn't started or available. If it's not defined
     the default value of `required` is `true`.
+- `service`: name of the target service (defaults to mapping key). Value can be set by a variable.
```
laurazard, line 384: (I'd add the example using this key under this line)

laurazard:
Yeah I guess you're right 😅. I left a couple comments just about examples, but otherwise LGTM

jhrotko (approved):

shershnevsergey:
Any news here?

ndeloof:
@shershnevsergey my bad, forgot I started working on this

ndeloof:
Closing as obsolete

PR #410, opened 2023-08-24
added build:network information

### LINKED ISSUE #409: `Network` attribute in `Build` missing
It seems that the `build` documentation does not contain information regarding the `network` attribute. 
You can set it during build times.

This issue has been discussed before in another repo (and closed), and information has been added to the official website doc (since v.3.4) :
https://github.com/docker/compose/issues/7911
https://docs.docker.com/compose/compose-file/compose-file-v3/#network

Made a PR regarding this, so that this doc is in sync with the website information. Wrote the text, character by character.

0xheartcode:

Appending information about `network` during `build`. You can define any network (or set it to none) during the build process for the `RUN` commands.
This information is already on the website.

Fixes #409


milas (approved):
LGTM - This is definitely supported by Compose for both the classic builder and BuildKit and was part of the v3 spec, so I think this was an oversight that this was missing.

https://github.com/docker/compose/blob/bd32ed1454450ed2b62f6f926e985757453f0e09/pkg/compose/build.go#L408

ndeloof (approved):

PR #416, opened 2023-09-06
Tiny typo fix

alfiedotwtf:
Found small duplications while reading the spec.

### DIFF
--- a/02-model.md
+++ b/02-model.md
@@ -16,7 +16,7 @@ A [secret](09-secrets.md) is a specific flavor of configuration data for sensiti
 >
 > With volumes, configs and secrets you can have a simple declaration at the top-level and then add more platform-specific information at the service level.
 
-A project is an individual deployment of an application specification on a platform. A project's name, set with the the top-level [`name`](04-version-and-name.md) attribute, is used to group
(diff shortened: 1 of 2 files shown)


laurazard (approved):

laurazard:
@alfiedotwtf  it looks like your commit message is missing a DCO sign-off, causing the DCO check to fail.

We require all commit messages to have a `Signed-off-by` line with your name and e-mail (see ["Sign your work"](https://github.com/moby/moby/blob/v20.10.11/CONTRIBUTING.md#sign-your-work), which looks something like:

```
Signed-off-by: YourFirsName YourLastName <yourname@example.org>
```

You can find some instructions on how to amend your commits to include the sign-off line in the [output of the DCO check](https://github.com/compose-spec/compose-spec/pull/416/checks?check_run_id=16531388502).

Sorry for the hassle, and thanks again for the contribution :)

alfiedotwtf:
@laurazard: Ah cool, no problem. I've amended the commit. Please let me know if you need me to do anything else :)

... as an added thought though, something the project may want to discuss on the mailing list - since my commit was GPG signed, that signed-off line may just be duplicating things?

milas (approved):

Issue #346, opened 2023-05-06
Configs inline content


bagratte:
**What is the problem you're trying to solve**
Declare config content directly in compose file, inline in `configs` top-level element.

**Describe the solution you'd like**
```yaml
name: config-inline
services:
  service:
    image: ubuntu
    entrypoint: cat
    command: /config.txt
    configs:
      - config.txt
configs:
  config.txt:
    data: |
      strawberry fields forever
```
```
$ docker compose up
[+] Running 2/2
 ✔ Network config-inline_default      Created              0.1s 
 ✔ Container config-inline-service-1  Created              0.1s 
Attaching to config-inline-service-1
config-inline-service-1  | strawberry fields forever
config-inline-service-1 exited with code 0
```
It might also be worth having base64 support.
```yaml
name: config-inline
services:
  service:
    image: ubuntu
    entrypoint: cat
    command: /config.txt
    configs:
      - config.txt
configs:
  config.txt:
    data_base64: |
      c3RyYXdiZXJyeSBmaWVsZHMgZm9yZXZlcgo=
```
```
$ docker compose up
[+] Running 2/2
 ✔ Network config-inline_default      Created              0.1s 
 ✔ Container config-inline-service-1  Created              0.1s 
Attaching to config-inline-service-1
config-inline-service-1  | strawberry fields forever
config-inline-service-1 exited with code 0
```

srstsavage:
File ownership/permissions probably would need to be handled as well? And it would probably make sense to either put configs in some predefined path like `/run/config/<configname>` or specify a path. But otherwise big +1.

```
name: config-inline
services:
  service:
    image: ubuntu
    entrypoint: cat
    command: /run/config/config.txt
    configs:
      - config.txt
configs:
  config.txt:
    uid: 1000
    gid: 1050
    mode: 0400
    data: |
      strawberry fields forever
```

or

```
name: config-inline
services:
  service:
    image: ubuntu
    entrypoint: cat
    command: /configs/config.txt
    configs:
      - config.txt:/configs/config.txt
configs:
  config.txt:
    uid: 1000
    gid: 1050
    mode: 0400
    data: |
      strawberry fields forever
```

dmckernanacv:
I like this idea but I think the top-level block should be `files` instead of `configs` as it could technically be any file, not just for configuration.

[...]

PR #429, opened 2023-10-31
introduce inline content for configs

ndeloof:
Introduce support for inlined content when defining a configuration file

Fixes https://github.com/compose-spec/compose-spec/issues/346

### DIFF
--- a/08-configs.md
+++ b/08-configs.md
@@ -13,6 +13,8 @@ By default, the config:
 The top-level `configs` declaration defines or references configuration data that is granted to services in your Compose application. The source of the config is either `file` or `external`.
 
 - `file`: The config is created with the contents of the file at the specified path.
+- `environment`: The config content is created with the value of an environment variable.
+- `content`: The content is created with the inlined value.
 - `external`: If set to true, `external` specifies that this config has already been created. Compose does not
   attempt to create it, and if it does not exist, an error occurs.
 - `name`: The name of the config object in the container engine to look up. This field can be used to
@@ -39,6 +41,22 @@ configs:
 ```
 
 ### Example 2
+
+`<project_name>_app_config` is created when the application is deployed,
+by registering the inlined content as the configuration data. This comes with the
+benefits Compose will infer variables when creating the config, which allows to
+adjust content according to service configuration:
+
+```yml
+configs:
+  app_config:
+    content: |
+      debug=${DEBUG}
+      spring.application.admin.enabled=${DEBUG}
+      spring.application.name=${COMPOSE_PROJECT_NAME}
+```
+
+### Example 3
 
 External configs lookup can also use a distinct key by specifying a `name`. 
 
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -764,6 +764,8 @@
       "type": "object",
       "properties": {
         "name": {"type": "string"},
+        "content": {"type": "string"},
+        "environment": {"type": "string"},
         "file": {"type": "string"},
         "external": {
           "type": ["boolean", "object"],
--- a/spec.md
+++ b/spec.md
@@ -2368,6 +2368,8 @@ By default, the config:
 The top-level `configs` declaration defines or references configuration data that is granted to services in your Compose application. The source of the config is either `file` or `external`.
 
 - `file`: The config is created with the contents of the file at the specified path.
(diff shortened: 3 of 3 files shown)


08-configs.md
```diff
@@ -13,6 +13,8 @@ By default, the config:
 The top-level `configs` declaration defines or references configuration data that is granted to services in your Compose application. The source of the config is either `file` or `external`.
 
 - `file`: The config is created with the contents of the file at the specified path.
+- `environment`: The config is created with the value of an environment variable.
```
glours, line 16:
```suggestion
- `environment`: The config content is created with the value of an environment variable.
```

glours (approved):
Just as nit to make things clearer

Bilge:
How does one benefit from this change? With the latest Docker Desktop 4.25.2 (129061), I get `configs.fpm Additional property content is not allowed`.

polarathene:
> `configs.fpm Additional property content is not allowed`.

I also tried this since it's published on the docs (_[`configs` top-level](https://docs.docker.com/compose/compose-file/08-configs/) page, service [`configs` attribute](https://docs.docker.com/compose/compose-file/05-services/#configs)_) and there is no min version description, so I assumed it was already available at some point.

I looked through the [release notes](https://docs.docker.com/compose/release-notes) and **found nothing** (_not too fun with fairly generic keywords `configs` / `content`_).

Then I tried to track down the source from the docs:
- There is no direct link for contributions on the rendered docs page, but it takes you to `docker/docs` repo.
- While some of the compose docs are there, this particular one was missing. Searched the repo and found [hugo config ref](https://github.com/docker/docs/blob/c47c4a4ba18a686a3ac2b353c2aec452aa55bbaa/hugo.yaml#L282C13-L299).
- Now at `compose-spec/compose-spec`, found the [docs page content and `git blame`](https://github.com/compose-spec/compose-spec/blame/ca5c01413d11fed0241f7726e03f6d4540ef3bb7/08-configs.md#L17) to find out it was merged in Nov 7th here.
  - `compose-go 1.20.1` (Nov 14th) [released with the feature](https://github.com/compose-spec/compose-go/releases/tag/v1.20.1).
  - [`docker/compose` upgraded to that dep](https://github.com/docker/compose/pull/11187) on the same day. Then [released v2.23.1](https://github.com/docker/compose/releases/tag/v2.23.1) on Nov 16th.
    - The release notes don't mention the feature, nor does the [Docker Compose docs release notes page](https://docs.docker.com/compose/release-notes/#2231).

That was a bit of a mission 😅 Didn't realize the feature was that new. I'm on v2.21.0 (Aug 30th).

I was on 4.23.0 (120376), but now also upgraded to the latest version is 4.25.2 (129061), but unfortunately that's not new enough 😆 

---

# TL;DR

> How does one benefit from this change?

[Docker Desktop 4.25.0](https://docs.docker.com/desktop/release-notes/#4250) (released Oct 26th) provides Docker Compose 2.23.0 (`docker compose version`).

While 4.25.2 (Nov 21st) probably could have bumped the compose version, we'll probably have to **wait a little bit longer** for a new release to handle that.

selivan:
@ndeloof 

I would like to suggest upgrading compose specification version for changes like this one.

Just now I've ran into this situation:
- I write `docker-compose.yml` according to official docs
- I specify the latest available `version: "3.8"`
- `docker compose` does not complain about incompatibility with this version
- And yet I get an error `Additional property content is not allowed`

That is confusing.

ndeloof:
> I specify the latest available version: "3.8"

`version` is obsolete with compose-spec, it is ignored. Compose spec isn't versioned _by design_. The compose (tool) version you're using will let you know it can't parse your compose file if you use a recent addition.

> docker compose does not complain about incompatibility with this version
And yet I get an error Additional property content is not allowed

This is exactly what "complain about incompatibility" is about. Maybe we should add "_please upgrade docker compose_" to this message.

selivan:
@ndeloof 

Unfortunately, there is no way to distinguish between _"please upgrade docker compose"_ and _"you mistyped a keyword or used it on incorrect nesting level"_.

> `version` is obsolete with compose-spec, it is ignored.

That is sad 😢 

Look at terraform and terraform providers, for example. Each version explicitly has it's own specification. As long as you follow that specification for selected version - you will not run into compatibility issues.

PR #430, opened 2023-11-03
un-deprecate scale

ndeloof:
This resurrect `scale` as a legitimate attribute to be set, as a simpler way to define replicas vs `deploy` section

### DIFF
--- a/05-services.md
+++ b/05-services.md
@@ -1495,9 +1495,8 @@ web:
 
 ## scale
 
-_DEPRECATED: use [deploy/replicas](deploy.md#replicas)_
-
 `scale` specifies the default number of containers to deploy for this service.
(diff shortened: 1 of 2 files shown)


ndeloof:
implemented by https://github.com/compose-spec/compose-go/commit/6e38069a2b68327d5059e9e3cd8733f58fd6571c

05-services.md
```diff
@@ -1495,9 +1495,8 @@ web:
 
 ## scale
 
-_DEPRECATED: use [deploy/replicas](deploy.md#replicas)_
-
 `scale` specifies the default number of containers to deploy for this service.
+When both are set, `scale` must be consistent with [deploy.replicas](deploy.md#replicas).
```
aevesdocker, line 1499:
```suggestion
When both are set, `scale` must be consistent with the `replicas` attribute in the [Deploy Specification](deploy.md#replicas).
```

glours (approved):

PR #432, opened 2023-11-08
clarify default values for ports declaration

ndeloof:

Fix compose spec regarding ports definition default value, according to de-facto implementation by `docker run -p ...` as [docker/go-connections](https://github.com/docker/go-connections)

Fixes https://github.com/compose-spec/compose-go/issues/458

### DIFF
--- a/05-services.md
+++ b/05-services.md
@@ -1369,7 +1369,7 @@ If host IP is not set, it binds to all network interfaces. Ports can be either a
 value or a range. Host and container must use equivalent ranges.
 
 Either specify both ports (`HOST:CONTAINER`), or just the container port. In the latter case,
-Compose automatically allocates any unassigned port of the host.
+the container runtime automatically allocates any unassigned port of the host.
 
 `HOST:CONTAINER` should always be specified as a (quoted) string, to avoid conflicts
 with [yaml base-60 float](https://yaml.org/type/float.html).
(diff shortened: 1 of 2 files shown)


05-services.md
```diff
@@ -1340,7 +1340,7 @@ If host IP is not set, it binds to all network interfaces. Ports can be either a
 value or a range. Host and container must use equivalent ranges.
 
 Either specify both ports (`HOST:CONTAINER`), or just the container port. In the latter case,
-Compose automatically allocates any unassigned port of the host.
+container runtime automatically allocates any unassigned port of the host.
```
aevesdocker, line 1343:
```suggestion
the container runtime automatically allocates any unassigned port of the host.
```

05-services.md
```diff
@@ -1373,8 +1373,8 @@ expressed in the short form.
 - `target`: The container port
 - `published`: The publicly exposed port. It is defined as a string and can be set as a range using syntax `start-end`. It means the actual port is assigned a remaining available port, within the set range.
 - `host_ip`: The Host IP mapping, unspecified means all network interfaces (`0.0.0.0`).
-- `protocol`: The port protocol (`tcp` or `udp`), unspecified means any protocol.
-- `mode`: `host`: For publishing a host port on each node, or `ingress` for a port to be load balanced.
+- `protocol`: The port protocol (`tcp` or `udp`). Defaults to `tcp`.
+- `mode`: `host`: For publishing a host port on each node, or `ingress` for a port to be load balanced. Defaults to `ingress`
```
aevesdocker, line 1377:
```suggestion
- `mode`: `host`: For publishing a host port on each node, or `ingress` for a port to be load balanced. Defaults to `ingress`.
```

aevesdocker (commented):
2 small pedantic comments from me

lionello:
Defaulting to "ingress" feels weird to me. I'd rather we spec if as "host" by default and fix the implementation.

laurazard (approved):
LGTM

ndeloof:
@lionello there's actually no implementation relying on this attribute, as it was introduced for docker swarm which doesn't use the compose specification. On swarm-disabled docker engine this has no effect

PR #435, opened 2023-11-08
Add a network-scoped mac_address field

akerouanton:

##### Add a network-scoped mac_address field

Since https://github.com/moby/moby/pull/45905 has been merged (will be part of the upcoming v25 release), it's now possible to specify an endpoint-specific MAC address. Moreover, setting the container-wide MAC address will return a warning.

### DIFF
--- a/05-services.md
+++ b/05-services.md
@@ -1216,6 +1216,10 @@ networks:
     driver: bridge
 ```
 
+#### mac_address
+
+`mac_address` sets the MAC address used by the service container when connecting to this particular network.
+
 #### priority
 
 `priority` indicates in which order Compose connects the service’s containers to its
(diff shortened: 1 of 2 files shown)


05-services.md
```diff
@@ -1243,7 +1247,7 @@ networks:
 
 ### mac_address
 
-`mac_address` sets a MAC address for the service container.
+_DEPRECATED: use [networks.mac_address](#mac_address)_
```
glours, line 1250: We can't mark it as deprecated if it's not also deprecated on the Docker CLI either

glours (changes requested):

glours:
You also need to update the `compose-spec.json` according to your changes

ndeloof (approved):

glours:
@akerouanton I forgot to mention, you need to run `make spec` to generate the `spec.md` with the changes you added

glours (approved):

akerouanton:
@glours should be good now.

PR #439, opened 2023-11-16
fix - fixes definition/development properties

stevendaniels:
**Fixes definition/development properties**:

Several properties for watch were left outside of `items`, which was causing errors when trying to import the spec to cue.

I was trying to import the spec into cue using the folowing commands:

```
curl -sL https://raw.githubusercontent.com/compose-spec/compose-spec/master/schema/compose-spec.json > compose-spec.json && \
cue import -f --outfile - -l '#ComposeSpec:' --package 'compose' compose-spec.json   
```

Which caused the following error:
```
constraint not allowed because type object is excluded:
    ./compose-spec.json:464:11
```
Moving the orphaned attributes into "items" fixed the issue for me.

Fixes #

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -459,11 +459,11 @@
               "path": {"type": "string"},
               "action": {"type": "string", "enum": ["rebuild", "sync", "sync+restart"]},
               "target": {"type": "string"}
-            }
-          },
-          "required": ["path", "action"],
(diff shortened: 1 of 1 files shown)


ndeloof (approved):

ndeloof:
For legal reasons contributors must sign-off commits when they contribute this repo. Please amend your commit and force-push your branch

PR #440, opened 2023-11-19
Fix an error in `dns_search` description

yumemio:

This PR fixes an error in the section describing [the `dns_search` configuration](https://github.com/compose-spec/compose-spec/blob/b0e5a16eaf4438ec9e8a8f25ee8fc0074c3d790c/05-services.md?plain=1#L472-L474). The current description erroneously refers to `dns` instead of `dns_search`.

N/A

### DIFF
--- a/05-services.md
+++ b/05-services.md
@@ -471,7 +471,7 @@ dns_opt:
 
 ### dns_search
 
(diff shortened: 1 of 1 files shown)


glours (approved):

glours:
good catch, thanks 🙏

PR #443, opened 2023-11-21
Permit '=' separator and '[ipv6]' in 'extra_hosts'

robmry:

Update the `extra_hosts` element to align with changes in 'docker/cli' and 'docker/buildx' that make it easier to specify IPv6 addresses in the `--add-host` option (by permitting `host=ip` in addition to `host:ip`, and allowing square brackets around the address).

For example:

```
extra_hosts:
    - "somehost=162.242.195.82"
    - "myhostv6=::1"
    - "anotherhostv6=[2001:4860:4860::8888]"
```

_My intention is to allow brackets around an address in the long/mapping form of `extra_hosts` as well, because it's unambiguous and consistent with the short form - but it's a slightly unusual thing to do, because the brackets are only normally used in a URI or when there's a port number - so, haven't added an example to the docs. But, to note it here, it'll look like ..._

```
extra_hosts:
    myhostv6: "[::1]"
```

Fixes docker/cli#4648

### DIFF
--- a/05-services.md
+++ b/05-services.md
@@ -850,12 +850,28 @@ external_links:
 `extra_hosts` adds hostname mappings to the container network interface configuration (`/etc/hosts` for Linux).
 
 ### Short syntax
-Short syntax uses plain strings in a list. Values must set hostname and IP address for additional hosts in the form of `HOSTNAME:IP`.
+Short syntax uses plain strings in a list. Values must set hostname and IP address for additional hosts in the form of `HOSTNAME=IP`.
+
+```yml
+extra_hosts:
+  - "somehost=162.242.195.82"
+  - "otherhost=50.31.209.229"
+  - "myhostv6=::1"
+```
+
+IPv6 addresses can be enclosed in square brackets, for example:
+
+```yml
+extra_hosts:
+  - "myhostv6=[::1]"
+```
+
+The separator `=` is preferred, but `:` can also be used. For example:
 
 ```yml
 extra_hosts:
   - "somehost:162.242.195.82"
-  - "otherhost:50.31.209.229"
+  - "myhostv6:::1"
 ```
 
 ### Long syntax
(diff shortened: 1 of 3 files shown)


05-services.md
```diff
@@ -850,12 +850,23 @@ external_links:
...
+Short syntax uses plain strings in a list. Values must set hostname and IP address for additional hosts in the form of `HOSTNAME=IP`.
+
+```yml
+extra_hosts:
+  - "somehost=162.242.195.82"
+  - "otherhost=50.31.209.229"
+  - "myhostv6=::1"
+```
+
+IPv6 addresses can be enclosed in square brackets and, for backward
```
ndeloof, line 862: is this "backward compatibility" as use of `:` is deprecated in docker/cli and user is warned (then https://github.com/compose-spec/compose-go/pull/499 should also do), or is this just an alternate (better?) syntax ? In such case, both should be documented, and maybe `=` recommended to prevent ambiguities

ndeloof (approved):
LGTM
just a question on `:` separator status: is this deprecated, or just we recommend use of `=` to avoid ambiguities ?

05-services.md
robmry: Ok, got it - thank you. So, instead of ...

> IPv6 addresses can be enclosed in square brackets and, for backward compatibility, the separator `:` can be used. For example:
> 
> extra_hosts:
>   - "somehost:162.242.195.82"
>   - "otherhost:50.31.209.229"
>   - "myhostv6:[::1]"

How about this? ...

> IPv6 addresses can be enclosed in square brackets, for example:
> 
> extra_hosts:
>   - "myhost=[::1]"
>
> The separator `=` is preferred, but `:` can also be used. For example:
> 
> extra_hosts:
>   - "somehost:162.242.195.82"
>   - "myhostv6:::1"

ndeloof (approved):

glours (approved):
LGTM

aevesdocker (approved):
LGTM!

dvdksn (approved):
thanks @robmry

PR #459, opened 2024-02-01
fix confusion between networks and network_mode

ndeloof:
`networks` section has description for "host or none" which actually applies to network_mode in services, not network top level element.

### DIFF
--- a/06-networks.md
+++ b/06-networks.md
@@ -66,46 +66,10 @@ driver is not available on the platform.
 ```yml
 networks:
   db-data:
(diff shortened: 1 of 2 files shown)


glours (approved):
LGTM

ndeloof:
AFAIK default installation has support for `bridge` (default) but also `macvlan` and `overlay`, and also allow use of direct `host` networking. This section should just refer to https://docs.docker.com/network/drivers/

PR #470, opened 2024-02-28
remove restriction for `extends` to refer to another resource

ndeloof:
for legacy reasons, `extends` was documented to forbid any reference to another service. Same consistency issue applies to volumes/networks/.. without such a constraint to be applied, and consistency check during parsing allows to discover misconfigurations. This PR removes this requirement, while documenting the need for compose file author to check resources exists in the extending model

see https://github.com/docker/compose/issues/11544

### DIFF
--- a/05-services.md
+++ b/05-services.md
@@ -660,15 +660,15 @@ extends:
 
 ### Restrictions
 
-The following restrictions apply to the service being referenced:
+Service being referenced by `extends` can have dependency declared on other resources. Typically it can have an explicit `volumes` declaration.
+`extends` then will not import the target volume definition in the extending compose model, it is Compose file author responsibility to define
+an equivalent resource for the extended service to be consistent. Compose will check a resource with referenced ID exists in the Compose model
 
-- Services that have dependencies on other services cannot be used as a base. Therefore, any key
-  that introduces a dependency on another service is incompatible with `extends`. The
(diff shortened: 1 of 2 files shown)


05-services.md
```diff
@@ -660,15 +660,15 @@ extends:
 
 ### Restrictions
 
-The following restrictions apply to the service being referenced:
+Service being referenced by `extends` can have dependency declared on other resources. Typically it can have an explicit `volumes` declaration.
+`extends` then will not import the traget volume definition in the extending compose model, it is Compose file author responsibility to define
```
glours, line 664:
```suggestion
`extends` then will not import the target volume definition in the extending compose model, it is Compose file author responsibility to define
```

05-services.md
```diff
@@ -660,15 +660,15 @@ extends:
...
+an equivalent resource for the extended service to be consistent. Compose will check a resource with referenced ID exist in the Compose model
```
glours, line 665:
```suggestion
an equivalent resource for the extended service to be consistent. Compose will check a resource with referenced ID exists in the Compose model
```

glours (approved):
LGTM
few typos to fix before merging

glours:
IHMO we should wait a little bit before merging, be sure `compose-go` is up to date and a Compose release coming soon

PR #472, opened 2024-03-14
introduce app_protocol to document level 7 protocol exposed by port

ndeloof:
Comparable to https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol
could be used for compose to offer hints on exposed ports (actual UX to be defined):
```
$ docker compose port
NAME                        PUBLISHED
test-web-1                http://localhost:8080
```

or as links in a GUI reporting compose app status


glours (approved):
LGTM

PR #479, opened 2024-03-31
Add missing "app_protocol" property

### LINKED ISSUE #478: Missing app_protocol property of "port" object in spec
The [documentation](https://docs.docker.com/compose/compose-file/05-services/#long-syntax-3) lists "app_protocol" as a property for service ports. This spec is missing that property.

I'd like for the property to be added.

ssnailed:

This property can be found in the [documentation](https://docs.docker.com/compose/compose-file/05-services/#long-syntax-3) but not in this spec.

Fixes #478

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -327,7 +327,8 @@
                   "host_ip": {"type": "string"},
                   "target": {"type": "integer"},
                   "published": {"type": ["string", "integer"]},
-                  "protocol": {"type": "string"}
+                  "protocol": {"type": "string"},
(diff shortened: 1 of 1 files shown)


ndeloof (approved):
my bad, forgot to update schema while writing https://github.com/compose-spec/compose-spec/pull/472

ndeloof:
For legal reason we require commits to use `--signoff`, please amend your commit and force push so we can merge

ssnailed:
My apologies, the commit is signed off now.

PR #493, opened 2024-04-30
Fix compose schema according to Draft7

jhrotko:
This PR fixes the compose schema, which was not previously valid. I changed the spec to Draft07.

Some of the string formats I needed to add a regex pattern, since some of them were only available in 2019 json schema spec. 

Also added tests to validate that the schema compiles, and some cases related to the added changes to make sure the behaviour is still the same according to the docker compose documentation.

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -1,6 +1,6 @@
 {
-  "$schema": "https://json-schema.org/draft/2019-09/schema#",
-  "id": "compose_spec.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "compose_spec.json",
   "type": "object",
   "title": "Compose Specification",
   "description": "The Compose file is a YAML file defining a multi-containers based application.",
@@ -20,14 +20,13 @@
     "include": {
       "type": "array",
       "items": {
-        "type": "object",
         "$ref": "#/definitions/include"
       },
       "description": "compose sub-projects to be included."
     },
 
     "services": {
-      "id": "#/properties/services",
+      "$id": "#/properties/services",
       "type": "object",
       "patternProperties": {
         "^[a-zA-Z0-9._-]+$": {
@@ -38,7 +37,7 @@
     },
 
     "networks": {
-      "id": "#/properties/networks",
+      "$id": "#/properties/networks",
       "type": "object",
       "patternProperties": {
         "^[a-zA-Z0-9._-]+$": {
@@ -48,7 +47,7 @@
     },
 
     "volumes": {
-      "id": "#/properties/volumes",
+      "$id": "#/properties/volumes",
       "type": "object",
       "patternProperties": {
         "^[a-zA-Z0-9._-]+$": {
@@ -59,7 +58,7 @@
     },
 
     "secrets": {
-      "id": "#/properties/secrets",
+      "$id": "#/properties/secrets",
       "type": "object",
       "patternProperties": {
         "^[a-zA-Z0-9._-]+$": {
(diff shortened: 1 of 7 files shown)


schema/compose-spec.json
```diff
@@ -165,12 +165,15 @@
...
-        "cpu_quota": {"type": ["number", "string"]},
-        "cpu_period": {"type": ["number", "string"]},
-        "cpu_rt_period": {"type": ["number", "string"]},
-        "cpu_rt_runtime": {"type": ["number", "string"]},
-        "cpus": {"type": ["number", "string"]},
+        "cpu_shares": {"anyOf":[{"type":"number"},{"type":"string"}]},
+        "cpu_quota": {"anyOf":[{"type":"number"},{"type":"string"}]},
+        "cpu_period": {"anyOf":[{"type":"number"},{"type":"string"}]},
+        "cpu_rt_period": {"anyOf":[{"type":"number"},{"type":"string"}]},
+        "cpu_rt_runtime": {"anyOf":[{"type":"number"},{"type":"string"}]},
```
milas, line 172: I believe these should be `oneOf`. (But in practice, I don't think it makes a difference, since it's not possible for something to fulfill both types simultaneously)

milas: (I think this applies to all the `anyOf`s in the file, but I won't comment on each one!)

tests/compose-spec_test.go
```diff
@@ -0,0 +1,227 @@
...
+			spec: createComposeSpecService(`
+        "ports": [
+            "3000",
+            "3000-3005",
+            "8000:8000",
+            "9090-9091:8080-8081",
+            "49100:22",
+            "8000-9000:80",
+            "127.0.0.1:8001:8001",
+            "127.0.0.1:5000-5010:5000-5010",
```
milas, line 124: Probably need a few IPv6 test cases here, which I think might fail the current regex

(but I assume engine supports)

tests/compose-spec_test.go
```diff
@@ -0,0 +1,227 @@
...
+
+			var data interface{}
+
+			// NOTE: make sure in array test element case does not have a ',' in the last element
+			err := json.Unmarshal([]byte(tt.spec), &data)
+			assert.NilError(t, err)
+
+			err = sch.Validate(data)
+			if !tt.assertError {
+				assert.NilError(t, err)
```
milas, line 219: You can avoid the `gotest.tools/v3` dependency by doing direct calls to `t.Fatalf` here (like the "should have thrown error for spec" assertion)

tests/go.mod
```diff
@@ -0,0 +1,10 @@
+module compose-spec/tests
+
+go 1.22.0
+
+require (
+	github.com/santhosh-tekuri/jsonschema/v5 v5.3.1
```
milas, line 6: This is a different library than we use in compose-go, right? Is that one not flexible enough for these tests?

milas: Also I wonder if we should have a GHA workflow to run `go test ./tests/...`

milas (commented):
Overall this LGTM!

A few comments for potential polish/clean-up, but I think this is pretty much ready to go

jhrotko:
Can someone give me permissions in this repo? :')

jhrotko:
Added a ci.yml Test step

ndeloof:
The reference for compose-spec.json is hosted on compose-go repository, this one is just a copy

jhrotko:
> The reference for compose-spec.json is hosted on compose-go repository, this one is just a copy (see https://github.com/compose-spec/compose-spec/blob/master/.github/workflows/sync.yml)

@ndeloof I am not understanding what you are suggesting. What does this imply?

ndeloof:
@jhrotko changes to json schema must take place on compose-go repo, and will be sync here by automation (we need to keep this copy as IDEs rely on this github URL)

PR #507, opened 2024-05-26
Fix deploy.placement constraints/preferences docs

### LINKED ISSUE #506: `deploy.placement` `constraints` and `placement` documentation does not match `compose-spec.json`
As an example, I want to declare a swarm service that, as a `preference`, prefers to be on a `worker` node.

The compose docs ([github](https://github.com/compose-spec/compose-spec/blob/a8751e853209baf5f57ae191a3ebd0cf491b45b3/deploy.md#placement), [webdocs](https://docs.docker.com/compose/compose-file/deploy/#placement)) clearly state that each of:
- `deploy.placement.constraints`
- `deploy.placement.preferences`

> It can be set either by a list or a map with string values.

However, in implementation, this is untrue, and the actual schema supported more closely matches what's described in the swarm services [documentation](https://docs.docker.com/engine/swarm/services/#control-service-placement).

The spec should be changed to support the documentation, or the documentation should be changed to accurately reflect what's supported in the spec.

In the likely scenario that there are practical implementation differences between the processing of `constraints` vs `preferences`, changing the documentation seems easier and more important.

**Additional context**
I'm able to practically demonstrate that the documentation is incorrect, as far as the reference implementation is concerned.
Relevant/matching issue on `docker/compose`: https://github.com/docker/compose/issues/11853

PaarthShah:
This PR is to fix the docs to match `compose-spec.json`'s AND `docker compose`'s implementation for each of:
-  `deploy.placement.constraints`
- `deploy.placement.preferences`

The docs have been modified to match the practical implementation of the spec as-is, including its usage for docker swarm: https://docs.docker.com/engine/swarm/services/#control-service-placement

Fixes #506

Renders https://github.com/docker/compose/issues/11853 irrelevant

### DIFF
--- a/deploy.md
+++ b/deploy.md
@@ -65,40 +65,48 @@ services:
 
 #### constraints
 
-`constraints` defines a required property the platform's node must fulfill to run the service container. It can be set either
-by a list or a map with string values.
+`constraints` defines a required property the platform's node must fulfill to run the service container.
+It must be set by a list with string values.
+
+Values are checked against node properties using an equality operator (`==` or `!=`).
+
+In the following example, the service can only runs on nodes that:
+- have the worker role
+- have the `region` label set to `east`
+- do not have a `disktype` of `hdd`
+
 
 ```yml
-deploy:
-  placement:
-    constraints:
-      - disktype=ssd
-```
-
-```yml
-deploy:
-  placement:
-    constraints:
-      disktype: ssd
+services:
+  frontend:
+    image: example/backend
+    deploy:
+      placement:
+        constraints:
+          - node.role == worker
+          - node.labels.region == east
+          - node.labels.disktype != hdd
 ```
 
 #### preferences
 
-`preferences` defines a property the platform's node should fulfill to run service container. It can be set either
-by a list or a map with string values.
+`preferences` defines an algorithm and property the platform's node should fulfill to run the service container.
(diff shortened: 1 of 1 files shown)


PaarthShah:
Ah I see I got somewhat ninja'ed by https://github.com/compose-spec/compose-spec/commit/20ebf93c4fca53ab0628b27d4c707c81639a6890

ndeloof:
As swarm does not apply the compose specification, I wonder this is relevant to rely on swarm docs to define those attributes in the compose specification. Better keep them with a fuzzy definition IMHO

PaarthShah:
@ndeloof I'm generally okay with it, but I think we can do better. I'd have to compare/contrast my PR with it properly (it's bit a little bit)

As far as implementation goes, I'm wondering if https://github.com/docker/cli/pull/4863#issuecomment-2132120246 is a good example at all, or whether it's purely hypothetical at this point.

And in the general case, I'd cite that the compose-spec is intended to be a superset of docker compose 2 and 3:

https://docs.docker.com/compose/compose-file/14-include/

> Legacy versions 2.x and 3.x of the Compose file format were merged into the Compose Specification.

Therefore I expect everything in compose file format 3 to be de-facto transferred to the compose spec, or explicitly marked for deprecation + removal when the implementation of something better is available.

I vote for the former because of the existence of PRs like the above that will provide implementation.

PR #515, opened 2024-07-02
Update compose-spec.json: Allow strings as includes

jonathanazulay:
Validation tooling warns about includes using strings directly. See attachment.
It is however allowed, which is already specified on the $ref definition. I think the additional `type: "object"` just overrides it.

Skärmavbild 2024-07-02 kl  12 49 30

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -20,7 +20,6 @@
     "include": {
       "type": "array",
       "items": {
-        "type": "object",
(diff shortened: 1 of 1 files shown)


jhrotko:
hello @jonathanazulay ! the source of truth is in `compose-go` repo. You should apply your changes inside the definition [`#/definitions/include`](https://github.com/compose-spec/compose-go/blob/main/schema/compose-spec.json#L630)

PR #530, opened 2024-09-15
IPv6 addresses can be in brackets for services.ports

k9withabone:

Specifies that IPv6 addresses in the `ports` short syntax can be enclosed in square brackets. This is already mentioned for the `extra_hosts` attribute, but not for ports. The reference library (compose-go) already supports IPv6 addresses with or without square brackets. This commit documents that fact and adds a couple of examples.

N/A, minor documentation change/clarification.


ndeloof (approved):

ndeloof:
ping

Issue #534, opened 2024-09-24
Invalid fields declared for the 'development/watch' array 


AnaMManolache:
**What is the problem you're trying to solve**
The definition of 'development/watch' contains 2 extra fields which are invalid. 'Watch' is an array type, so  'additionalProperties' and 'patternProperties' shouldn't be used in this case.
In jsonschema,  'additionalProperties' and 'patternProperties' are keywords used for objects, they are not applicable to arrays.

This is causing issues when converting this jsonschema to other formats, like cue (https://cuelang.org).

**Describe the solution you'd like**
These fields are already specified for the items of this array, so these extra fields can be removed.

**Additional context**

AnaMManolache:
It works now! Thanks!

PR #535, opened 2024-09-24
Update compose-spec.json: remove invalid fields from array

AnaMManolache:

The 'additionalProperties' and 'patternProperties' are being declared for an array type, which is invalid. They are also declared for the items of the array, so they can be removed. Their presence is causing issues converting this jsonschema to its cuelang representation.

Fixes [534.](https://github.com/compose-spec/compose-spec/issues/534)

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -483,9 +483,7 @@
             },
             "additionalProperties": false,
             "patternProperties": {"^x-": {}}
-          },
-          "additionalProperties": false,
-          "patternProperties": {"^x-": {}}
+          }
         }
       }
     },


glours:
Hello @AnaMManolache 
Thanks for you contribution, the main `compose-spec.json` is located [in the `compose-go` library](https://github.com/compose-spec/compose-go/blob/main/schema/compose-spec.json), so you should open your PR in `compose-go` repo

glours:
replace by https://github.com/compose-spec/compose-go/pull/684

glours:
@AnaMManolache we updated this repo with your changes done in compose-go, you can try to regenerate your cuelang representation

PR #541, opened 2024-11-13
clarify condition compose will parse string as variable

ndeloof:

the exact implementation of compose variable parser should not be defined in the spec, still we can clarify a bit condition a `$` sign won't trigger a parser error.

Fixes https://github.com/docker/compose/issues/12287


glours (approved):
Sounds good to me 👍

PR #543, opened 2024-11-18
document supported tmpfs options

aevesdocker:

Fixes #

### DIFF
--- a/05-services.md
+++ b/05-services.md
@@ -1755,6 +1755,7 @@ parameters (sysctls) at runtime](https://docs.docker.com/engine/reference/comman
 
 ```yml
 tmpfs: 
+ - <path>
  - <path>:<options>
 ```
 
@@ -1771,6 +1772,7 @@ services:
   app:
     tmpfs:
       - /data:mode=755,uid=1009,gid=1009
+      - /run
 ```
 
 ## tty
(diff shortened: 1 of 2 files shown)


05-services.md
```diff
@@ -1754,13 +1754,23 @@ parameters (sysctls) at runtime](https://docs.docker.com/engine/reference/comman
 `tmpfs` mounts a temporary file system inside the container. It can be a single value or a list.
 
 ```yml
-tmpfs: /run
+tmpfs: 
+ - <path>:<options>
```
ndeloof, line 1758: as `options` is optional, better document example with and without options

glours: Good point @ndeloof, I agree with having examples with & without options

05-services.md
```diff
@@ -1754,13 +1754,23 @@ parameters (sysctls) at runtime](https://docs.docker.com/engine/reference/comman
...
 ```
 
+- <path>: The path inside the container where the tmpfs will be mounted.
+- <options>: Comma-separated list of options for the tmpfs mount.
+
+Available options:
```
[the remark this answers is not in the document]
ndeloof: yes, that's ok as the spec exposes daemon and Linux flags in many places

PR #546, opened 2024-12-02
label_file

ndeloof:
see https://github.com/compose-spec/compose-go/pull/713


ndeloof:
@aevesdocker yes this will be part of next release (added label accordingly)

glours (approved):

PR #551, opened 2024-12-17
Add network option enable_ipv4 (for moby 28.0)

robmry:

**_Update description of option enable_ipv6_**

- "enables IPv6 networking" -> "enables IPv6 address assignment"
  - aligns with updated CLI help text
- Replace ref to an engine docs "step 4" that doesn't exist with an inline example.

**_Add option enable_ipv4_**

The plan is to hook this up to the network `CreateOptions.EnableIPv4` field that will be added in moby 28.0, https://github.com/moby/moby/pull/48271.

### DIFF
--- a/06-networks.md
+++ b/06-networks.md
@@ -95,6 +95,19 @@ networks:
     attachable: true
 ```
 
+### enable_ipv4
+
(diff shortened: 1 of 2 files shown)


akerouanton (approved):

robmry:
> Added an update to `schema/compose-spec.json`.

Removed again - apparently a bot will sort it out.

Issue #554, opened 2025-01-07
Drop restriction related to `extends` regarding other references to other resources


ndeloof:
`extends` was introduced by https://github.com/docker/compose/pull/1088 with explicit constraint references to other services (`volumes_from` and others) are prohibited, as `extends` _only_ import service definition in the compose model, without any form of transitivity. This decision has been taken after initial experiment to support transitive import/copy resulted in various challenges.

In the meantime, Compose v2 has been implemented and didn't fully reproduce this restriction, while this was noted in the compose specification (which is basically compose v1 documentation made a spec). This was fixed by https://github.com/compose-spec/compose-go/pull/718, but then we noticed some users have been relying on this "feature".

I suggest we remove this restriction, and let user extend services even those refer to other resources. This:
1. restores backward compatibility with earlier compose v2 release
2. doesn't break compose v1 compatibility as a valide compose v1 model would not allow those, we just relax the rule
3. requires end-user to be responsible for such resources being declared in the compose model. Considering compose is a well establish tool, maybe we don't need as much baby sitting as the original authors have to

Vigilans:
Agree on dropping the restriction. Since compose has complex mechanisms for composing final configuration model:
* Multiple config files by -f
* Extends
* Include

Even if `depends_on` is filtered from `extends`, it can still be imported by providing another configuration file via `-f`.

Therefore, all semantic processing should only start after configs loading and merging is done, i.e. a comprehensive model is available now. From this perspective `extends` are just used like a module snippet for reuse, it's surprising and not reasonable to exclude certain fields.

PR #555, opened 2025-01-09
drop extends restrictions

ndeloof:

Fixes https://github.com/compose-spec/compose-spec/issues/554


glours (approved):
LGTM

PR #557, opened 2025-01-09
clarify `command` syntax vs Dockerfile comparison

ndeloof:

Fixes https://github.com/docker/compose/issues/12403

### DIFF
--- a/05-services.md
+++ b/05-services.md
@@ -190,7 +190,19 @@ cgroup_parent: m-executor-abcd
 command: bundle exec thin -p 3000
 ```
 
-The value can also be a list, in a manner similar to [Dockerfile](https://docs.docker.com/engine/reference/builder/#cmd):
+> **Note**
+> 
+> Unlike the `CMD` instruction of an image, the [shell-form syntax](https://docs.docker.com/reference/dockerfile/#shell-form) for `command` 
+> does not implicitly run in the context of the [`SHELL` instruction](https://docs.docker.com/reference/dockerfile/#shell).
+> 
+> If you expect the command to rely on features of a shell environment such as environment variables, then ensure the command is run within a shell:
+> 
+> ```yaml
+> command: /bin/sh -c 'echo "hello $$HOSTNAME"'
+> ```
+
+The value can also be a list, in a manner similar to the [exec-form syntax](https://docs.docker.com/reference/dockerfile/#exec-form) 
+used by [Dockerfile](https://docs.docker.com/engine/reference/#cmd).
 
 ```yaml
 command: [ "bundle", "exec", "thin", "-p", "3000" ]
--- a/spec.md
+++ b/spec.md
@@ -403,7 +403,19 @@ cgroup_parent: m-executor-abcd
 command: bundle exec thin -p 3000
 ```
 
-The value can also be a list, in a manner similar to [Dockerfile](https://docs.docker.com/engine/reference/builder/#cmd):
+> **Note**
+> 
+> Unlike the `CMD` instruction of an image, the [shell-form syntax](https://docs.docker.com/reference/dockerfile/#shell-form) for `command` 
+> does not implicitly run in the context of the [`SHELL` instruction](https://docs.docker.com/reference/dockerfile/#shell).
+> 
+> If you expect the command to rely on features of a shell environment such as environment variables, then ensure the command is run within a shell:
+> 
+> ```yaml
+> command: /bin/sh -c 'echo "hello $$HOSTNAME"'
+> ```
+
+The value can also be a list, in a manner similar to the [exec-form syntax](https://docs.docker.com/reference/dockerfile/#exec-form) 
+used by [Dockerfile](https://docs.docker.com/engine/reference/#cmd).
 
 ```yaml
 command: [ "bundle", "exec", "thin", "-p", "3000" ]


05-services.md
```diff
@@ -190,7 +190,9 @@ cgroup_parent: m-executor-abcd
 command: bundle exec thin -p 3000
 ```
 
-The value can also be a list, in a manner similar to [Dockerfile](https://docs.docker.com/engine/reference/builder/#cmd):
+The value can also be a list, defining the exact command and arguments to be ran by container.
+This syntax can be usefull to disambiguate the syntax used by simpler form, when Compose fails 
+to correctly parse arguments.
```
polarathene, line 195:
```suggestion
> **Note**
>
> Unlike the `CMD` instruction of an image, the [shell-form syntax](https://docs.docker.com/reference/dockerfile/#shell-form) for `command` does not implicitly run in the context of the [`SHELL` instruction](https://docs.docker.com/reference/dockerfile/#shell).
> 
> If you expect the command to rely on features of a shell environment such as environment variables, then ensure the command is run within a shell:
> 
> ```yaml
> command: /bin/sh -c 'echo "hello $$HOSTNAME"'
> ```
>
> When the `entrypoint` (_or image `ENTRYPOINT`_) is configured to run the shell instead, to ensure `command` is processed correctly you should prefer the [exec-form syntax](https://docs.docker.com/reference/dockerfile/#exec-form) instead.

The value can also be a list, in a manner similar to [Dockerfile](https://docs.docker.com/engine/reference/#cmd).
```

If choosing the suggestion, it might be worth also linking to [interpolation](https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/) page for context on why it's `$$HOSTNAME` instead of `$HOSTNAME` to the reader, in case they are not aware of the syntax as it's not exactly easy to search up? 🤔 

I removed `builder/` from the link as it no longer seems to be used when resolving the URL.

---

<details>
<summary>Not relevant to this PR</summary>

The list example that follows could potentially also demonstrate a YAML feature that can be useful when wanting to pass multiple commands to a shell via multiple lines instead of `&&` chaining a long string. Or perhaps that's off-topic / niche?

```yaml
command: /bin/sh -c 'echo hello && echo world'
command: ["/bin/sh", "-c", "echo hello && echo world"]

# More appropriate for a sequence of commands in exec form
# if not running a separate shell script within the image
command:
  - /bin/sh
  - -c
  - |
    echo hello
    echo world
```

There's similar YAML syntax for long commands with many args to an executable in shell-form, but allowing you to define it as multiple lines without any trailing `\n` (_I recall that being relevant at some point_).

</details>

spec.md
```diff
@@ -403,7 +403,9 @@ cgroup_parent: m-executor-abcd
 command: bundle exec thin -p 3000
 ```
 
-The value can also be a list, in a manner similar to [Dockerfile](https://docs.docker.com/engine/reference/builder/#cmd):
+The value can also be a list, defining the exact command and arguments to be ran by container.
+This syntax can be usefull to disambiguate the syntax used by simpler form, when Compose fails 
+to correctly parse arguments.
```
polarathene, line 408: If not using the revised suggestion, there are two typos to correct:

```suggestion
The value can also be a list, defining the exact command and arguments to be ran by the container.
This syntax can be useful to disambiguate the syntax used by simpler form, when Compose fails 
to correctly parse arguments.
```

polarathene (commented):
I don't think the current addition properly conveys the concern. I've provided a suggestion that has more context at explains the difference better to the reader, although it doesn't get too technical.

<details>
<summary>Original note</summary>

One noticeable "bug" is an exec-form entrypoint will take only the first component (space delimited?) of a shell-form command and discard the rest. I think this only occurs when it's determined to be an executable? (_for example, `command: uname -m`_)
- The equivalent outside of Compose:
  - Via `CMD` injects the `SHELL` which ignored the command if ENTRYPOINT exec-form also launched a shell, due to shell nesting?
  - Via `docker run` behaved the same as Compose `command`.

That can be avoided by ensuring `docker run` or Compose `command` ensure the value `uname -m` is provided as a single string. `command: "'uname -m"'`

</details>

**UPDATE:** Covered the original note concern [here](https://github.com/docker/compose/issues/12403#issuecomment-2597237191). The `docker run` docs don't really communicate this either, but command is parsed as exec-form, while presumably only Compose is capable of the alternative list/exec-form syntax via YAML (_`docker run` is less flexible for `--entrypoint` in that sense_). Perhaps the docs should better emphasize that distinction, that `entrypoint` / `command` are always parsed to the equivalent of exec-form?

ndeloof:
@polarathene I dropped references to entrypoint used to run a shell as this is a corner case and uncommon practice. Most users either run image with a plain command and no entrypoint, or have an entrypoint script which expect some parameters passed as command (see https://docs.docker.com/build/building/best-practices/#entrypoint)

polarathene:
Oh absolutely, I cited the `/bin/sh -c` entrypoint due to the original issue to explain why it was happening and expected, I don't think your original response really explained it there? 😅 

Then the follow-up when I realized myself that compose (and docker CLI) were always effectively exec-form. I had encountered problems in the past when setting command and wondering why some images the behaviour seemed different.

---

Instead of setting an image entrypoint to blank if I wanted to start it with a shell instead I was overriding with `/bin/sh for `docker run --rm -it`, which can't override an entrypoint to `/bin/sh -c` like Compose which I sometimes did for `compose.yaml` examples I'd make (_like using the `smallstep/step-ca` image to run a custom command that another service `depends_on`_).

It seemed easier to use `configs.x.content` to embed as a shell script instead and set that as the entrypoint, or use YAML list with `command` (_or [read the `entrypoint` docs](https://docs.docker.com/reference/compose-file/services/#entrypoint) to realize I should set `''` / `[]`_). It wasn't until responding in the associated issue that it clicked how to provide an actual single string value like the `CMD` shell-form is treated, due to `command` only processing the value like it is exec-form. In your original PR change here you referred to the YAML list form as a way to avoid ambiguity which sparked the curiosity to what the difference was 😅

PR #558, opened 2025-01-15
gpus

ndeloof:
see https://github.com/docker/compose/pull/12197#issuecomment-2588891336

### DIFF
--- a/05-services.md
+++ b/05-services.md
@@ -941,6 +941,29 @@ configuration, which means for Linux `/etc/hosts` get extra lines:
 ::1             myhostv6
 ```
 
+## gpus
+
+[![Compose NEXT RELEASE](https://img.shields.io/badge/compose-NEXT-blue?style=flat-square)](https://github.com/docker/compose/releases/NEXT)
+
+`gpus` specifies GPU devices to be allocated for container usage. This is equivalent to a [device request](deploy.md#devices) with
+an implicit `gpu` capability.
+
+```yaml
+services:
+  model:
+    gpus: 
+      - driver: 3dfx
+        count: 2
+```
+
+`gpus` also can ge set as string `all` to allocate all aviable GPU devices to container.
+
+```yaml
+services:
+  model:
+    gpus: all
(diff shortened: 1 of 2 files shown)


polarathene:
If this is still to be defined, would there be any reason to require the `- count: all` value vs just `gpus: all`?

My understanding was it was a convenience for the docker CLI, that configuring additional capabilities (_which nvidia has for more niche requirements_) would be done with the more verbose alternative.

Nope, I remembered wrong it's a list of string key/value pairs:
- https://docs.docker.com/reference/cli/docker/container/run/#gpus
- https://docs.docker.com/engine/containers/resource_constraints/#gpu

```bash
docker run -it --rm --gpus all ubuntu nvidia-smi
docker run -it --rm --gpus device=GPU-3a23c669-1f69-c64e-cf85-44e9b07e7a2a ubuntu nvidia-smi
docker run -it --rm --gpus '"device=0,2"' ubuntu nvidia-smi
docker run --gpus 'all,capabilities=utility' --rm ubuntu nvidia-smi
```

I guess that could translate to `all` or a list of devices equivalent to `deploy.resources.reservations.devices[]`. The PR diff implies `count` will map `all` to `-1`?

---

Other nitpick is AFAIK these are used for Docker Compose docs, while the spec is agnostic. Docker CLI for `--gpus` implies it's nvidia only, and given the move towards CDI it may remain that way?

I don't know what the equivalent of Compose Deploy looks like the Docker CLI, my understanding was it was meant to be more lower-level/granular (_no nvidia assumption for the GPU? Allowing support for AMD/Intel_). Likewise with CDI, that distinction matters less and perhaps the spec/docs for `gpus` here should also raise awareness towards CDI / `device`? (_assuming `gpus` is to become potentially deprecated in future?_)

ndeloof:
@polarathene this is the current status of implementation by compose-go, I have plan to introduce such a shortcut notation in a follow-up PR

glours (approved):

PR #560, opened 2025-01-22
compose spec parity

aevesdocker:

PR aims to get more parity between Compose Specs in terms of descriptions and clarity of attributes. No meaningful content has been deleted, just moved around to get it alphabetised.

I have removed the `[Compose v2.22.0](https://github.com/docker/compose/releases/v2.22.0)` as we should keep this in the docs repo version and keep this version neutral. 

A few more bits to do but they need to be done in the docs repo.

Fixes #

### DIFF
--- a/05-services.md
+++ b/05-services.md
@@ -22,6 +22,8 @@ If not implemented the `deploy` section is ignored and the Compose file is still
 
 ## annotations
 
+[![Compose v2.20.0](https://img.shields.io/badge/compose-v2.20.0-blue?style=flat-square)](https://github.com/docker/compose/releases/v2.20.0)
+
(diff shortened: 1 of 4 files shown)


ndeloof:
> I have removed the [Compose v2.22.0](https://github.com/docker/compose/releases/v2.22.0) as we should keep this in the docs repo version and keep this version neutral.

Why ? docker/compose being the reference implementation, even if the original intent for the spec was to be vendor neutral, it makes perfect sense we have those badges

glours (commented):
As Nicolas, I think we should keep the badge with the Compose version implementing
Otherwise LGTM, thanks @aevesdocker 🙏

ndeloof (approved):

glours (approved):

PR #566, opened 2025-01-31
schema: remove unrecognized keywords

rogpeppe:
I noticed in passing that some of the entries inside the JSON Schema did not do anything:
- under 2019-09 `id` does not mean anything (as of draft-06, it's [spelled `$id`](https://json-schema.org/draft-06/json-schema-release-notes)), and there's no need to define explicit IDs for internal references.
- there is no recongized "expose" format
- the `constraints` definition is never referenced and has no constraints (`service` is not a valid keyword).

I've chosen a top level `$id` URI that works, but this is arbitrary and could be replaced by something less github-specific: although not a requirement, it's nice if this is a working URL.

I've also added a newline at the end of the file because that's reasonably standard.

This new schema should be functionally identical to the original.

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2019-09/schema#",
-  "id": "compose_spec.json",
+  "$id": "https://raw.githubusercontent.com/compose-spec/compose-spec/HEAD/schema/compose-spec.json",
   "type": "object",
   "title": "Compose Specification",
   "description": "The Compose file is a YAML file defining a multi-containers based application.",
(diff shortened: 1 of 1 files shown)


ndeloof:
schema reference is hosted on https://github.com/compose-spec/compose-go/
please propose your fix there

schema/compose-spec.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2019-09/schema#",
-  "id": "compose_spec.json",
+  "$id": "https://raw.githubusercontent.com/compose-spec/compose-spec/HEAD/schema/compose-spec.json",
```
ndeloof, line 3: :+1
fixed earlier by https://github.com/compose-spec/compose-go/commit/25fe17966316466351cf15d5c09c888e28a760e6 but unfortunately mixed other changes we had to revert

ndeloof:
I cherry-picked your changes into https://github.com/compose-spec/compose-go/pull/733

PR #572, opened 2025-02-24
introduce refresh pull policy

ndeloof:

See https://github.com/docker/compose/issues/12563


glours (approved):
👍

Issue #138, opened 2021-02-04
Description of fields in compose-spec.json


mik-laj:
Hello,

I would like to suggest that you add description fields that describe individual fields so you don't need to look for the description of fields in specific documentation, but the documentation was available in tools that can use descriptions fields.

For example: I have these specifications loaded in Jetbrains IDE.
<img width="469" alt="Screenshot 2021-02-04 at 16 47 05" src="https://user-images.githubusercontent.com/12058428/106918684-b68d1d00-6709-11eb-9c80-9f2dae386417.png">
It makes sure my files are automatically linted while working on them and I get instant feedback.
<img width="453" alt="Screenshot 2021-02-04 at 16 56 43" src="https://user-images.githubusercontent.com/12058428/106918972-fc49e580-6709-11eb-926a-0007ca3032df.png">
I can also easily go to the specification if I press CTRL+field.
<img width="642" alt="Screenshot 2021-02-04 at 16 58 26" src="https://user-images.githubusercontent.com/12058428/106919228-3c10cd00-670a-11eb-910c-5dcae0f1cf6d.png">
I would like to be able to read the description as well, but unfortunately it is not included in the specification. Other specs have descriptions which makes my life easier. This is what it looks like in the Workflow or Github Action specification. 
<img width="422" alt="Screenshot 2021-02-04 at 17 02 02" src="https://user-images.githubusercontent.com/12058428/106919792-c0fbe680-670a-11eb-85a3-53c54ef0820a.png">
<img width="1241" alt="Screenshot 2021-02-04 at 17 02 10" src="https://user-images.githubusercontent.com/12058428/106919796-c22d1380-670a-11eb-9b8a-4862b4997306.png">

As a next step, we can also generate user documentation and publish it on the website but it's an extra effort and it's enough for me that the specification has descriptions. See: https://github.com/compose-spec/compose-spec/issues/115

Best regards,
Kamil Breguła

EricHripko:
Hi @mik-laj 👋 Thank you for suggesting this idea - I definitely agree that it's worth providing short descriptions for attributes directly in the schema.

bwateratmsft:
[...]

PR #581, opened 2025-03-19
Add descriptive documentation to compose schema properties

testA113:
Improved schema documentation by adding descriptive text to various properties in the JSON schema. 

This enhances DX by providing clear explanations of what each property does without requiring users to constantly refer to external documentation.

These changes address issue #138.

The descriptions come from sane descriptions found in https://github.com/microsoft/compose-language-service/blob/7a74283ddb866988fed86241f461a657f0aee6d0/src/service/providers/KeyHoverProvider.ts#L63-L184

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -8,7 +8,7 @@
   "properties": {
     "version": {
       "type": "string",
-      "description": "declared for backward compatibility, ignored."
+      "description": "The version of the Docker Compose document"
     },
 
     "name": {
@@ -31,7 +31,8 @@
           "$ref": "#/definitions/service"
         }
       },
-      "additionalProperties": false
+      "additionalProperties": false,
+      "description": "The services in your project"
     },
 
     "networks": {
@@ -40,7 +41,8 @@
         "^[a-zA-Z0-9._-]+$": {
           "$ref": "#/definitions/network"
         }
-      }
+      },
+      "description": "Networks that are shared among multiple services"
     },
 
     "volumes": {
@@ -50,7 +52,8 @@
           "$ref": "#/definitions/volume"
         }
       },
-      "additionalProperties": false
+      "additionalProperties": false,
+      "description": "Named volumes that are shared among multiple services"
     },
 
     "secrets": {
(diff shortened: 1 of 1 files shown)


schema/compose-spec.json
```diff
@@ -8,7 +8,7 @@
   "properties": {
     "version": {
       "type": "string",
-      "description": "declared for backward compatibility, ignored."
+      "description": "The version of the Docker Compose document"
```
ndeloof, line 11: Please, keep a warning so user **don't** use this anymore

ndeloof (approved):
LGTM, could be later improved

ndeloof (changes requested):
`version` attribute must be documented as legacy, deprecated

schema/compose-spec.json
testA113: Sure thing, updated

testA113: Sure thing, updated

testA113:
Thankyou for the review @ndeloof! Let me know if you need any other changes

glours:
This PR should be done in the https://github.com/compose-spec/compose-go repository, no?
The version of the `compose-spec.json` file in this repo is just a copy of the compose-go one

rcjsuen:
> This PR should be done in the https://github.com/compose-spec/compose-go repository, no? The version of the `compose-spec.json` file in this repo is just a copy of the compose-go one

https://github.com/compose-spec/compose-spec/blob/d9b3470a67e91dd5abbc8c4e06e0c1982b89a010/Makefile#L22-L24

If that is the case then we should add a `README.md` in the `schema` folder to educate users about this. I don't think this is clear to people looking at this repository unless they start looking at the code.

testA113:
> Improved schema documentation by adding descriptive text to various properties in the JSON schema.
> 
> This enhances DX by providing clear explanations of what each property does without requiring users to constantly refer to external documentation.
> 
> These changes address issue #138.
> 
> The descriptions come from sane descriptions found in https://github.com/microsoft/compose-language-service/blob/7a74283ddb866988fed86241f461a657f0aee6d0/src/service/providers/KeyHoverProvider.ts#L63-L184

Thanks for taking a look @glours! I've applied the same changes in the PR for the compose-go repo you mentioned [here](https://github.com/compose-spec/compose-go/pull/776). I'll close this PR and let's continue the review there

PR #585, opened 2025-04-09
introduce volume.type=image

ndeloof:

document https://github.com/docker/compose/pull/12725


glours (approved):
LGTM

PR #411, opened 2023-08-25
codespell: add config, github workflow to avoid common typos; also  fix some typos

yarikoptic:

PR #600, opened 2025-05-21
Add codespell support (config, workflow to detect/not fix) and make it fix some typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

This is redone
- #411 

which I hope would not wait as long to accumulate conflicts ;-)


yarikoptic:
done

PR #604, opened 2025-06-26
introduce models

ndeloof:
documentation for https://github.com/compose-spec/compose-go/pull/799

cc @xenoscopic as I'm not confident enough with this topic to offer relevant descriptions :)

### DIFF
--- a/05-services.md
+++ b/05-services.md
@@ -1249,6 +1249,35 @@ There is a performance penalty for applications that swap memory to disk often.
 - If `memswap_limit` is unset, and `memory` is set, the container can use as much swap as the `memory` setting, if the host container has swap memory configured. For instance, if `memory`="300m" and `memswap_limit` is not set, the container can use 600m in total of memory and swap.
 - If `memswap_limit` is explicitly set to -1, the container is allowed to use unlimited swap, up to the amount available on the host system.
 
+## models
+
+`models` defines the AI models that the service containers use at runtime, referencing entries under the
+[top-level `models` key](models.md)
+
+```yaml
+services:
+  short_syntax:
+    image: app
+    models:
+      - my_model
+
+  long_syntax:
+    image: app
+    models:
+      my_model:
+        endpoint_var: MODEL_URL
+        model_var: MODEL
+
+```
+
+When using the long syntax,
+- `endpoint_var` can be set to define the environment variable set in container for the model runner URL. 
+If not explicitly declared, variable name is computed from model key (*) with a `_URL` suffix. In previous example, `short_syntax` service
+will run with the `MY_MODEL_URL` environment variable set.
+- `model_var` can be set to define the environment variable set in container for the model identifier. If not explicitly declared, variable name is computed from model key (*)
+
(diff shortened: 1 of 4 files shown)


schema/compose-spec.json
```diff
@@ -728,6 +764,10 @@
           "$ref": "#/definitions/ulimits",
           "description": "Override the default ulimits for a container."
         },
+        "use_api_socket": {
```
xenoscopic, line 767: Is this an unrelated change?

schema/compose-spec.json
```diff
@@ -1522,6 +1562,31 @@
...
+          "type": "string",
+          "description": "Language Model to run."
+        },
+        "context_size": {
+          "type": "integer"
+        },
+        "runtime_flags": {
+          "type": "string",
+          "description": "Raw runtime flags to pass to the inference engine."
+        }
```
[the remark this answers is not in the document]
xenoscopic: It's currently a string in the provider, but I'd prefer it be an array in the spec.  I think we can adjust the backend to support either.

spec.md
```diff
@@ -1460,6 +1460,30 @@ There is a performance penalty for applications that swap memory to disk often.
 - If `memswap_limit` is unset, and `memory` is set, the container can use as much swap as the `memory` setting, if the host container has swap memory configured. For instance, if `memory`="300m" and `memswap_limit` is not set, the container can use 600m in total of memory and swap.
 - If `memswap_limit` is explicitly set to -1, the container is allowed to use unlimited swap, up to the amount available on the host system.
 
+## models
```
xenoscopic, line 1463: I think we should put an explicit statement that a model artifact will be pulled and available to use via the inference endpoint APIs **_before_** any dependent service starts.

schema/compose-spec.json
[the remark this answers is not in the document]
xenoscopic: Let's go with a list of strings.

05-services.md
```diff
@@ -1249,6 +1249,30 @@ There is a performance penalty for applications that swap memory to disk often.
...
+
+  long_syntax:
+    image: app
+    models:
+      my_model:
+        variable: MY_MODEL
+
+```
+
+When using the long syntax, `variable` can be set to define the environment variable set in container for model runner URL.
```
xenoscopic, line 1272: I think we confirmed offline, but `endpoint_var` might be a more descriptive name.

schema/compose-spec.json
```diff
@@ -1522,6 +1562,32 @@
...
+          "description": "AI Model to run."
+        },
+        "context_size": {
+          "type": "integer"
+        },
+        "runtime_flags": {
+          "type": "array",
+          "items": {"type": "string"},
+          "description": "Raw runtime flags to pass to the inference engine."
+        }
```
glours, line 1584: Should we propose also `endpoint_var` as attribut for main model definition, so it will allow user to define custom env variable for the model endpoint once for all the services and let them override it per service, wdyt?

glours (commented):
1 question and a nit to fix 😉

glours (approved):
LGTM

PR #606, opened 2025-07-08
Expand models examples to show some common configurations

ilopezluna:
Provides developers with practical examples of how to configure AI models for different use cases

### DIFF
--- a/models.md
+++ b/models.md
@@ -11,13 +11,28 @@ services:
     models:
       - ai_model
 
-
 models:
   ai_model:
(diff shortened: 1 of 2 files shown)


kiview (approved):

ndeloof (approved):

kiview:
I think that is a fair point @ndeloof.
This would mean weaving them into https://docs.docker.com/ai/compose/models-and-compose/ 

cc @ArthurFlag

PR #612, opened 2025-09-01
Add details about the latest tag for PullPolicy Missing

telaoumatenyanis:

The spec does not give details about the behavior of `PullPolicyMissing` when the `latest` tag is used. From the current docker compose implementation, it looks like `latest` is always pulled, regardless of the usage of PullPolicyMissing. I'm proposing to align the specs (and the docs) with the behavior to reduce confusion.

Source: https://github.com/docker/compose/blob/04b8ac5fe4b9801953d1e6bff7465f4fe8bff377/pkg/compose/pull.go#L167

### DIFF
--- a/05-services.md
+++ b/05-services.md
@@ -1766,6 +1766,7 @@ events.
 * `missing`: Compose pulls the image only if it's not available in the platform cache.
    This is the default option if you are not also using the [Compose Build Specification](build.md).
   `if_not_present` is considered an alias for this value for backward compatibility.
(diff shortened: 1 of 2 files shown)


ndeloof:
This behavior has been introduced by https://github.com/docker/compose/pull/9781

As `docker run` does not behave this way, I don't thing this is legitimate for Compose to pretend it better manage `latest` tag. As we now have support for time-based refresh pull policy, better align with the Docker CLI IMHO

telaoumatenyanis:
Thanks Nicolas. Amended commit with sign-off

PR #630, opened 2026-02-10
Add `chown` option for bind and volume mounts

k9withabone:

The ability to recursively chown a volume's source is useful when running rootless containers, especially when using automatic user namespace provisioning. Podman supports this with the `U` volume option. See the `--volume` section of the [**podman-run**(1) documentation](https://docs.podman.io/en/v5.7.1/markdown/podman-run.1.html#volume-v-source-volume-host-dir-container-dir-options) for details.

Added the `U` option to the short service volume syntax.

Added the `bind.chown` and `volume.chown` options to the long service volume syntax.

### DIFF
--- a/05-services.md
+++ b/05-services.md
@@ -2089,6 +2089,7 @@ The short syntax uses a single string with colon-separated values to specify a v
   - `ro`: Read-only access.
   - `z`: SELinux option indicating that the bind mount host content is shared among multiple containers.
   - `Z`: SELinux option indicating that the bind mount host content is private and unshared for other containers.
+  - `U`: Recursively chown the source volume using the UID and GID of the container.
 
 > **Note**
 >
@@ -2122,9 +2123,11 @@ expressed in the short form.
   - `propagation`: The propagation mode used for the bind.
   - `create_host_path`: Creates a directory at the source path on the host if it doesn't exist. Defaults to `true`.
   - `selinux`: The SELinux re-labeling option `z` (shared) or `Z` (private)
+  - `chown`: Flag to recursively chown the source using the UID and GID of the container.
 - `volume`: Configures additional volume options:
   - `nocopy`: Flag to disable copying of data from a container when a volume is created.
   - `subpath`: Path inside a volume to mount instead of the volume root.
+  - `chown`: Flag to recursively chown the source volume using the UID and GID of the container.
 - `tmpfs`: Configures additional tmpfs options:
   - `size`: The size for the tmpfs mount in bytes (either numeric or as bytes unit).
   - `mode`: The file mode for the tmpfs mount as Unix permission bits as an octal number. [![Compose v2.14.0](https://img.shields.io/badge/compose-v2.14.0-blue?style=flat-square)](https://github.com/docker/compose/releases/v2.14.0)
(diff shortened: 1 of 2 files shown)


ndeloof:
While implemented by Podman, I don't like this proposal exposes the actual solution selected to `chown` the source for a bind mount, which has obvious side effects and can just prevent user to later access his own files 🤪.

While there's no out-of-the-box solution to manage UID mismatch between host and container, alternative approaches could be used, for example by creating a temporary idmapped mount (or FUSE bindfs) based on source folder so file ownership appears to be container's USER.

TheRealBecks:
@Luap99 I think I we need a perspective from the Podman team here. Currently the `U` mount flag is not part of the compose-spec and therefore hinders us (the Podlet maintainers) to guarantee (docker/podman) compose compatibility with Podman Quadlets.

Luap99:
> While there's no out-of-the-box solution to manage UID mismatch between host and container, alternative approaches could be used, for example by creating a temporary idmapped mount (or FUSE bindfs) based on source folder so file ownership appears to be container's USER.

Well podman also does have an `idmap` mount option too to automatically create an idmapped mount, but well of course that is only allowed when running as root so only rootful podman can use it.

For rootless podman users they need the chown way.



> While implemented by Podman, I don't like this proposal exposes the actual solution selected to chown the source for a bind mount, which has obvious side effects and can just prevent user to later access his own files 🤪.

Well AFAIK many container images work around that by doing the chown themselves as part of the entrypoint when they are started as root and only then drop to a users which is of course far worse then just starting the container with --user and letting the container manger chown it.


Generally I do no use compose and have no opinions on its spec, but why do the extra options need to be defined at all? Could you just not define it as arbitrary string/object and let the implementation define what options they actually support?

TheRealBecks:
At the end rootless containers running with Docker and Podman can run into this situation where `chown` is needed for the user to be able to access the files. Podman has implemented the `U` flag to fulfill the need. Quadlets (for Systemd) on the other hand also have this option to be Podman compatible.

Then there's the project [k9withabone/compose_spec_rs](https://github.com/k9withabone/compose_spec_rs) (will be containers/compose_spec_rs in the near future) that takes the `compose-spec` and builds a Rust module out of it that is widely used by several projects including the Podlet project.

Without this feature in compose-spec we can't add it to compose_spec_rs as it would provide a wrong feature set to other projects. @ndeloof I hope the last answers made it more clear to you why we think it's a good idea to add this option to the specs 🙂

glours:
@TheRealBecks @Luap99 I agree with @ndeloof and I’m also concerned about exposing novice users to a feature that could cause them to lose access to their mounted files.

As an interim solution while we work out the best way to address this properly, what would you think about introducing an extension attribute such as `x-chown` (or any other name you prefer), interpreted only by `compose_spec_rs`, to provide behavior equivalent to Podman’s `-U` option?

ndeloof:
I agree there's an obvious need to manage permission on bind mounts, but use of `chown` noun just reflects a specific implementation, that should not be exposed by the spec.

ndeloof:
> Well AFAIK many container images work around that by doing the chown themselves as part of the entrypoint.

You can use this pattern:
```Dockerfile
RUN mkdir /foo && chown user /foo
VOLUME /foo
```
when a container is created, volume is mounted with content/ownership declared from image. This is how volume ownership should be managed to be aligned with image uid/gid. This only applies to a fresh new volume, to prevent breaking other's usage of existing data

> Could you just not define it as arbitrary string/object and let the implementation define what options they actually support?

The reason this spec exists is to define attributes and what they are expected to provide at runtime. Implementation-specific option are well supported using `x-*` attributes, which are actually arbitrary string/object. Docker bake typically has been using those to offer advanced build options.

PR #642, opened 2026-04-17
introduce Jobs

### LINKED ISSUE #92: Proposal: Scheduled Jobs
This proposal is to gather community feedback on the idea of adding "scheduled jobs" to the compose-spec. 

## What are scheduled jobs?
A scheduled job is a task which is run at some defined interval (defined by a cron syntax). Unlike services, these executions of the container are ephemeral - they are spun up when it's time, and exit once they're complete. 

## What problem does this solve?
There are many problems that are well suited to being modeled as happening only periodically. Some examples are:

* Sending nightly emails
* Sending notifications to chatrooms
* Purging database records 
* Polling some data source
* Some batch/analytic job

The imagination is the limit on types of workloads that can be modeled as scheduled jobs!

## Is there a need for this?

There's certainly president. There are lots of third party tools which helps customers model their scheduled jobs via docker-compose. The [Ofelia](https://github.com/mcuadros/ofelia#readme) project, for example, exposes the schedule that a command should be run via labels on a service.

So another question would be, why not just follow this model - allow customers to set up services and tag them with some metadata? I think this model works well - but it forces people to write their scheduled jobs as services. While this works fine locally, as customers transition their compose apps to cloud, they'll be faced with the fact that services are often more expensive (operationally and financial) than a simple scheduled job. 
  
## What might it look like?

```yaml
services:
  web:
    build: ./web-app
    depends_on:
      - db
      - redis
  redis:
    image: redis
  db:
    image: postgres
jobs:
  email-sender:
    schedule: @daily
    build: ./email-sender
    command: bundle exec send-email
    environment:
      from: david@kohidave.dev
    depends_on:
      -db
```

ndeloof:

# Introduce jobs top-level element and a layered container specification

## Motivation

Services are long-running processes, but Compose users also need to define one-shot tasks — database migrations, batch processing, scheduled cleanups. A common hack is to declare those as "services" with a profile so they are excluded by default, and invoke them with `docker compose run`.
This PR introduces jobs as a first-class top-level element in the Compose specification to address this gap.

To avoid duplicating the full container configuration schema between element types, the shared configuration is extracted as a **layered specification**:

- **`container_spec`** — the attributes of a single container, shared by anything that runs one: services, jobs, and run-to-completion init containers (`pre_start` hooks);
- **`workload_spec`** — orchestrated-workload attributes (`build`, `depends_on`, `expose`, `healthcheck`, `ports`, `stdin_open`, `tty`), shared by services and jobs;
- service-only attributes (`attach`, `container_name`, `restart`, `scale`, `deploy`, `develop`, lifecycle hooks, …) stay in the services document.

As a direct consequence, `pre_start` hooks become a full container specification (`allOf [container_spec]` + `per_replica`) instead of a hand-curated 7-attribute subset: undeclared attributes are inherited from the service, declared ones override with the merge-file rules — the direction proposed in #656. Attributes excluded from inheritance by that proposal (`ports`, `healthcheck`, `depends_on`, `build`, …) are excluded *by construction*: they live in `workload_spec`, which hooks do not compose.

```yaml
jobs:
    db-migration:
      image: myapp:latest
      command: python manage.py migrate
      triggers:
        manual: true

    cleanup:
      image: busybox
      command: sh -c 'find /data -mtime +30 -delete'
      triggers:
        schedule: "0 3 * * *"
```

Note: this requires JSON Schema draft 2020-12 as proposed on https://github.com/compose-spec/compose-go/pull/864 so we can use `unevaluatedProperties` for composition by `allOf/$ref`

Fixes #92
Addresses #656 (schema side)


ndeloof:
cc @julienbreux

JulienBreux:
Very good, how can I help 🤗

glours:
@JulienBreux you’ll need to wait a little bit, this PR is only a draft for now. We still have some internal design work to finish first, as this needs to be supported on the Engine and CLI sides as well 😉

But we’re working on it 🎉

JulienBreux:
> @JulienBreux you’ll need to wait a little bit, this PR is only a draft for now. We still have some internal design work to finish first, as this needs to be supported on the Engine and CLI sides as well 😉
> 
> But we’re working on it 🎉

Thanks! Brilliant! That's a really good news.

PR #647, opened 2026-06-16
Add pre_start lifecycle hook for init containers

glours:
Introduce `pre_start` as a sequence of init containers run before a service starts. 
Each step runs to completion in declared order, in its own ephemeral container created after `ContainerCreate` but before `ContainerStart`, with the service container's mounts shared. A non-zero exit fails the bring-up of the service and its dependents.

This is the first-class replacement for the restart: "no" plus `service_completed_successfully` pattern.

Follow up of https://github.com/compose-spec/compose-spec/issues/11

### DIFF
--- a/05-services.md
+++ b/05-services.md
@@ -1687,6 +1687,45 @@ services:
           - FOO=BAR
 ```
 
+## pre_start
+
+`pre_start` defines a sequence of init containers to run before the service container is started. Each step runs to completion, in declared order, and the service container only starts once every step has exited 0. A non-zero exit fails the bring-up of the service and its dependents.
+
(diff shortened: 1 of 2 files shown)


ndeloof (approved):

psviderski:
It's really great to see the progress on init containers in docker compose and the spec. Awesome work! 👍 

I designed something very similar as a `x-pre_deploy` compose extension in Uncloud: https://uncloud.run/docs/guides/deployments/pre-deploy-hooks/
Will look into supporting/replacing it with `pre_start`. Is there any public discussion happening on this design and changes?

PR #652, opened 2026-07-06
Clarify mem_swappiness percentage range and effects

ygerlach:
Is the `mem_swappiness` parameter the same as the host parameter `/proc/sys/vm/swappiness` ? if so the current documentation is misleading. I tried to come up with a better text.

Fixes #


ygerlach:
i think this is still relevant

PR #654, opened 2026-08-03
Add link to Compose Bridge for Unified Defense Stack

willswire:

Adds a linked example of an implementation leveraging the spec for compose bridge


ndeloof (approved):
LGTM

By the way, do you know you could rely on https://docs.docker.com/compose/bridge/ and package model conversion as a transformer image ?

